PR #53, opened 2021-01-27
feat: concurrent deliverTx

jinsan-line:
Related with: https://github.com/line/link/issues/1170, https://github.com/line/tendermint/pull/170

## Description

To optimize performance, we need to increase concurrency. After implementing concurrent checkTx(#49) and recheckTx(#52), I'd like to implement concurrent deliverTx.

## Checklist:

- [ ] I followed the [contributing guidelines](https://github.com/line/link/blob/master/CONTRIBUTING.md).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,5 +47,5 @@ require (
 replace (
 	github.com/keybase/go-keychain => github.com/99designs/go-keychain v0.0.0-20191008050251-8e49817e8af4
 	github.com/tendermint/iavl => github.com/line/iavl v0.14.4-0.20210122121727-7c33089230bd
-	github.com/tendermint/tendermint => ../tendermint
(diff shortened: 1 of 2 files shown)


zmanian:
I would expect this implementation to introduce non-determimism and apphash failures. I expect that you need ReadWrite lists or some other mechanism to prevent transactions from concurrently accessing.

In prior design work Cosmos contributors have thoughts after putting access lists natively in the transaction format. Now some applications might want to have dynamic runtime access patterns so it would necessary to have optional support for ReadWrite lists which implies that only part of a block will contain transactions that can be processed in parallel.

egonspace:
Thank you for the great comment. This seems to be the first outside comment since opening our project, and we welcome it.

This PR was an experimental branch to determine whether performance was improved or not. The main does not include concurrent delivery Tx yet, and several issues need to be resolved. I'll listen carefully to your opinion.

Thank you so much!

PR #265, opened 2021-07-07
feat: Introduce sig block height for the new replay protection

egonspace:
## Description
Introduce `sig block height` to be added to tx body that is included to sign bytes.
And I removed account number concept.
This PR includes following features.
- sig block height field in tx body
- add the configuration `VALID_SIG_BLOCK_PREIOD` that limits how old the sig block height from current.
- removed account number
- linked empty tx of auth module to main app
- unknown account may send tx (it can pass the signature verification of ante handler)
- fix a bug that `--offline` flag of broadcast command does not work.

closes: https://github.com/line/lfb-sdk/issues/258

### How to use
```
simd tx bank send link1u02ad8y2k6c8y94fhc38wm9t2pvzr90t8wlecp link19wgf6ymq2ur6r59st95e04e49m69z4al4fc982 1stake --chain-id sim --keyring-backend test --sig-block-height 100
// Of course, it is possible to omit --sig-block-height here. If you omit it, then the node use current block height as that value
// But you should specify sig-block-height to use --generate-only or --offline flag.
```

### To test sending tx from unknown account
```
simd tx auth empty link1u02ad8y2k6c8y94fhc38wm9t2pvzr90t8wlecp --chain-id sim --keyring-backend test
// It's possible the account link1u02ad8y2k6c8y94fhc38wm9t2pvzr90t8wlecp is not on the chain at this moment.
```

### Configuration
We can configure `valid_sig_block_period` of auth param by modifying `genesis.json`
This configuration property limits how much older the value of the sig block height can be than the current block height.
The default value is 100.
```
"auth": {
      "params": {
        "max_memo_characters": "256",
        "tx_sig_limit": "7",
        "tx_size_cost_per_byte": "10",
        "sig_verify_cost_ed25519": "590",
        "sig_verify_cost_secp256k1": "1000",
        "valid_sig_block_period": "100"
      },
      "accounts": []
```
---

write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Code follows the [module structure standards](https://github.com/cosmos/cosmos-sdk/blob/master/docs/building-modules/structure.md).
- [x] Added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] Re-reviewed `Files changed` in the Github PR explorer
- [x] Review `Codecov Report` in the comment section below once CI passes


x/auth/ante/sigverify.go
```diff
@@ -269,18 +272,13 @@ func (svd *SigVerificationDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simu
...
-			ChainID:       chainID,
-			AccountNumber: accNum,
-			Sequence:      acc.GetSequence(),
+			ChainID:  chainID,
+			Sequence: acc.GetSequence(),
 		}
 
 		if !genesis {
-			sigKey := fmt.Sprintf("%d:%d", signerData.AccountNumber, signerData.Sequence)
+			sigKey := fmt.Sprintf("%s:%d", acc.GetAddress().String(), signerData.Sequence)
```
whylee259, line 281: Minor,
The sig cache can be overwritten with the same sigKeys.
```
on checkTX
TX1: sender A with seq 0 and sig a
cache: A:0=>a
TX2: remove account A 
TX3: sender A with seq 0 and sig b
cache: A:0=>b

on deliverTX
TX1: cache miss
TX2
TX3: cache hit
```

x/auth/ante/basic.go
```diff
@@ -200,3 +200,32 @@ func (txh TxTimeoutHeightDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
...
+func (txs TxSigBlockHeightDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if !simulate {
+		params := txs.ak.GetParams(ctx)
+		sbh := tx.GetSigBlockHeight()
+		current := uint64(ctx.BlockHeight())
+		validMin := uint64(0)
+		if current > params.ValidSigBlockPeriod {
+			validMin = current - params.ValidSigBlockPeriod
+		}
+		if sbh > uint64(ctx.BlockHeight()) || sbh < validMin {
```
whylee259, line 223: Does it exist any replay protection logic within `ValidSigBlockPeriod`?

egonspace: We will have to restrict the deletion of an account that signed ever within this `ValidSigBlockPeriod`. Although this implementation is not necessary because there is no account deletion now, it should be remembered when implementing account deletion in the future.

whylee259 (approved):
LGTM

iproudhon (approved):

x/auth/ante/basic.go
```diff
@@ -200,3 +200,32 @@ func (txh TxTimeoutHeightDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
...
+		ak: ak,
+	}
+}
+
+func (txs TxSigBlockHeightDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if !simulate {
+		params := txs.ak.GetParams(ctx)
+		sbh := tx.GetSigBlockHeight()
+		current := uint64(ctx.BlockHeight())
+		validMin := uint64(0)
```
brew0722, line 219: Is it ok even if block start height is not 1 by specifying `initalHeight`?

egonspace: For example, if start height is 50 and `ValidSigBlockPeriod` is 100 and current height is 80, then `sig block height` 40 is valid according to the code above. I think the `sig block height` lower than start height has no problem.

brew0722 (approved):

PR #281, opened 2021-07-13
perf: internal objects optimization

iproudhon:
perf: internal objects optimization
1. replaced cdc.(Un)MarshalInterface() by protobuf (Un)Marshal for internal objects: BaseAccount, Balance & Supply
2. PubKey in BaseAccount is secp256k1.PubKey, not interface, to avoid 'Any' type handling
3. address fields in the above objects and MsgSend (hence tx as well) are no longer strings, but sdk.AccAddress


iproudhon:
Some test files were not updated.

PR #287, opened 2021-07-14
perf: internal objects optimization

iproudhon:
For internal account, balance, supply objects, use protobuf marshal / unmarshal instead of much slower MarshalInterface. Also uses []byte for addresses instead of string in the above objects and MsgSend for performance as well.

write a little note why.


x/bank/keeper/msg_server.go
```diff
@@ -29,20 +29,14 @@ func (k msgServer) Send(goCtx context.Context, msg *types.MsgSend) (*types.MsgSe
...
 	}
-	from, err := sdk.AccAddressFromBech32(msg.FromAddress)
-	if err != nil {
-		return nil, err
-	}
-	to, err := sdk.AccAddressFromBech32(msg.ToAddress)
-	if err != nil {
-		return nil, err
-	}
+	from := sdk.AccAddress(msg.FromAddress)
```
brew0722, line 32: How does it work exist prefix validation by bech32?('link...')

egonspace: Client should assign binary address value to `Address` type of all messages.
We will have to think about this and discuss it.

brew0722: Ah, I temporally wrong things thought that the address bytes contained bech32 prefix bytes. I understood now.

iproudhon:
Further change is needed.

PR #320, opened 2021-09-09
perf: internal objects optimization (BaseAccount, Balance & Supply)

iproudhon:
* replaced cdc.(Un)MarshalInterface() with protobuf (Un)Marshal for
  internal objects: BaseAccount, Balance & Supply
* PubKey in BaseAccount has three seperate entries for publick keys to
  avoid MarshalInterface.
* custom json marshaler for BaseAccount to show a single pub key for
  json compatibility

---

write a little note why.

- [x] Targeted PR against correct branch (see [CONTRIBUTING.md](https://github.com/cosmos/cosmos-sdk/blob/master/CONTRIBUTING.md#pr-targeting))


x/auth/types/account.go
```diff
@@ -301,6 +336,41 @@ type AccountI interface {
 
 	// Ensure that account implements stringer
 	String() string
+
+	MarshalX() ([]byte, error)
+}
+
+func MarshalAccountX(cdc codec.BinaryMarshaler, acc AccountI) ([]byte, error) {
+	if bacc, ok := acc.(*BaseAccount); ok && bacc.MultisigPubKey == nil {
```
egonspace, line 344:
```suggestion
	if bacc.MultisigPubKey == nil {
		return acc.MarshalX()
	}
```
Isn't is possible?

x/auth/types/account.go
```diff
@@ -334,3 +404,121 @@ type GenesisAccount interface {
...
+}
+
+func (acc *BaseAccount) UnmarshalJSONPB(m *jsonpb.Unmarshaler, bz []byte) error {
+	var bi BaseAccountJSON
+
+	err := json.Unmarshal(bz, &bi)
+	if err != nil {
+		return err
+	}
+	/* TODO: do we need to validate address format here
```
egonspace, line 446: Yes. I think we'd better validate the address here. I don't think this part affects performance much.

x/auth/types/account.go
iproudhon: We don't know if acc is *BaseAccount, so we need to cast it first. AccountI can be one of *BaseAccount, *ModuleAccount or *GenesisAccount.

x/auth/types/account.go
iproudhon: The issue was we need to call three validators.

if err := sdk.ValidateAccAddress(bi.Address); err != nil {
    if err = sdk.ValidateValAddress(bi.Address); err != nil {
        if err = sdk.ValidateConsAdderss(bi.Address); err != nil {
            return err
        }
    }
}

Is there better way than this?

egonspace: Then let's do this. Only the address entered by the user is verified. It's probably msg handler.
And `Unmarshal` is not validate an address because it operates only on marshaled data.

egonspace:
Let's update `CHANGELOG`.

iproudhon:
Updated CHANGELOG

x/auth/types/account.go
iproudhon: Hence, I'll keep it as is.

egonspace (approved):
LGTM

proto/lfb/auth/v1beta1/auth.proto
```diff
@@ -1,6 +1,9 @@
 syntax = "proto3";
 package lfb.auth.v1beta1;
 
+import "lfb/crypto/ed25519/keys.proto";
+import "lfb/crypto/multisig/keys.proto";
+import "lfb/crypto/secp256k1/keys.proto";
 import "cosmos_proto/cosmos.proto";
 import "gogoproto/gogo.proto";
 import "google/protobuf/any.proto";
```
leesj9476, line 9: "google/protobuf/any.proto" seems to be unused, so is would be ok to delete it.

leesj9476 (approved):

proto/lfb/auth/v1beta1/auth.proto
egonspace: Applied

x/auth/types/account.go
```diff
@@ -301,6 +336,41 @@ type AccountI interface {
...
+	if len(bz) < sigLen {
+		return nil, fmt.Errorf("invalid data")
+	}
+	if bytes.Equal(bz[:sigLen], BaseAccountSig) {
+		acc := &BaseAccount{}
+		if err := acc.Unmarshal(bz[sigLen:]); err != nil {
+			return nil, err
+		}
+		return acc, nil
+	} else if bytes.Equal(bz[:sigLen], ModuleAccountSig) {
```
whylee259, line 364: There is no contract that `BaseAccountSig ` and `ModuleAccountSig ` are the same length.
This is likely to lead to mistakes later.

x/bank/keeper/send.go
```diff
@@ -266,7 +266,10 @@ func (k BaseSendKeeper) SetBalance(ctx sdk.Context, addr sdk.AccAddress, balance
 	balancesStore := prefix.NewStore(store, types.BalancesPrefix)
 	accountStore := prefix.NewStore(balancesStore, AddressToPrefixKey(addr))
 
-	bz := k.cdc.MustMarshalBinaryBare(&balance)
+	bz, err := balance.Marshal()
+	if err != nil {
+		return err
+	}
```
whylee259, line 272: Any reason to change this?
Do we ignore the legacy amino codec?
> https://github.com/line/lbm-sdk/blob/a18b7c8c01a77e5a1290a0a4600feda80ad3abce/codec/codec.go#L15-L16

iproudhon: It's done because there's measurable differences in performance between MarshalBinaryBare and protobuf's Marshal.

Yes, we'll ignore legacy codec as these are internal object serialization that's never going to be exposed.

x/auth/types/account.go
iproudhon: I agree. Will fix it with a separate check-in.

PR #347, opened 2021-10-06
feat: introduce `VoterSetCounter`

0Tech:
## Description

Introducing a concept "voter set counter" eases integration
with `ostracon`'s `VoterSet`.

---

write a little note why.


x/slashing/keeper/infractions.go
```diff
@@ -69,11 +70,11 @@ func (k Keeper) HandleValidatorSignature(ctx sdk.Context, addr cryptotypes.Addre
...
 	}
 
-	minHeight := signInfo.StartHeight + k.SignedBlocksWindow(ctx)
+	minVoterSetCount := k.SignedBlocksWindow(ctx) - 1
 	maxMissed := k.SignedBlocksWindow(ctx) - minSignedPerWindow
 
-	// if we are past the minimum height and the validator has missed too many blocks, punish them
-	if height > minHeight && signInfo.MissedBlocksCounter > maxMissed {
+	// if we have joined enough times to voter set and the validator has missed too many blocks, punish them
+	if voterSetCounter >= minVoterSetCount && signInfo.MissedBlocksCounter > maxMissed {
```
leesj9476, line 77: Why "voterSetCounter >= minVoterSetCount"?
I understood that anyone who voted less than minVoterSetCount times will not be punished.
Then, wouldn't it be "voterSetCounter > minVoterSetCount" or "minVoterSetCount := k.SignedBlocksWindow(ctx)" on line 73?

0Tech: It could be off-by-one bug, if I omit the equal sign because `voterSetCounter` counts from 0, which means the window would be filled at `voterSetCounter == k.SignedBlocksWindow(ctx) - 1`. Please correct me if my calculation was wrong.

0Tech: I've got your point. This may cause backward compatibility issue. They (cosmos-sdk) set the minimum height as `StartHeight + k.SignedBlocksWindow(ctx)` so effectively they punish only the validator whose active period exceeds `k.SignedBlocksWindow(ctx)` by 2. Weird but the truth is..

leesj9476: hmm.... that is very weird.... I understood about equal sign. 

My point is diffenent, a little bit.
After getting the signInfo.VoterSetCounter value from line 30, increment this value.
In the syntax comparing voterSetCounter on line 77, this value can be confusing because it is a value from the past, not current.


index := signInfo.voterSetCounter % k.SignedBlocksWindow(ctx)
signInfo.VoterSetCounter++  
voterSetCounter := signInfo.VoterSetCounter

If this code is edited like this, voterSetCounter store current value and the '-1' can be removed on line 73, which seems to be clear the confusion.

0Tech: I found the excess could be related to `ValidatorUpdateDelay`. The variable is hardcoded and its value is 1. Even if it is not the case, it seems like the comparison logic needs a patch because `ValidatorUpdateDelay` was not taken into account. @egonspace I want to discuss about this.

Issue #363, opened 2021-10-29
Empty response for invalid address on withdraw-all-rewards


egonspace:
## Summary
This is the feedback from the QA process of lbm v0.1.0-rc1.

simd responses empty for non-genesis account on `withdraw-all-rewards` command

```
simd tx distribution withdraw-all-rewards --from link1a35jrgnakgwarfc22c7wckh5hsn34r9r36v0a9 --keyring-backend test --home ~/.simapp/simapp0/ --chain-id sim
```

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #364, opened 2021-10-29
fix: force genOrBroadcastFn even when max-msgs != 0

0Tech:
## Description

If cli tx withdraw-all-rewards sends empty tx, the tx would be
rejected by validators because it has no msgs. However, the cli sends
empty tx only if one sets `--max-msgs=0` which means it does not split
the msgs. When you set `max-msgs` with any positive number, the cli
does not send any tx, which may confuse you because you cannot get the
feedback indicating the address has no delegations.
This patch addresses the problem, by forcing genOrBroadcastFn when the
number of total msgs generated is zero. It will provide the same user
experience with any `max-msgs`.

closes: #363 

---

write a little note why.

### DIFF
--- a/x/distribution/client/cli/tx.go
+++ b/x/distribution/client/cli/tx.go
@@ -53,13 +53,12 @@ func newSplitAndApply(
 	genOrBroadcastFn newGenerateOrBroadcastFunc, clientCtx client.Context,
 	fs *pflag.FlagSet, msgs []sdk.Msg, chunkSize int,
 ) error {
-
-	if chunkSize == 0 {
+	totalMessages := len(msgs)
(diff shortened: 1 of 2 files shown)


egonspace:
I think it's a good solution. However, it seems that the target branch should be `release/v0.43.x`. If so, it will be automatically applied in `main` branch by the github action.

dudong2 (approved):
LGTM

egonspace (approved):
LGTM

dudong2 (approved):

0Tech:
I have submitted the [corresponding PR](https://github.com/cosmos/cosmos-sdk/pull/10477) to cosmos-sdk. I will apply this PR after they respond.

Issue #384, opened 2021-11-24
Validator whitelist feature


egonspace:
## Summary
For the composition of the consortium chain, a validator whitelist feature is required so that only authorized accounts can register the validator.

For this, two functions are required.
- [staking] Check whitelist while it does `CreateValidator`
- [gov] following 3 new proposals
  - `ProposalAddToValidatorWhitelist`
  - `ProposalDeleteFromValidatorWhitelist`
  - `ProposalDisableValidatorWhitelist`: If it passed, the whitelist is deleted. It is required for changing the chain to public.
  - `ProposalEnableValidatorWhitelist`

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

egonspace:
I think a query for viewing the validator whitelist should also be implemented.

0Tech:
I think we can use same proposal for `ProposalDisableValidatorWhitelist` and `ProposalEnableValidatorWhitelist` with a boolean flag (e.g. enable), so if it is `true` it enables the feature, otherwise, it disables the feature.

egonspace:
> I think we can use same proposal for `ProposalDisableValidatorWhitelist` and `ProposalEnableValidatorWhitelist` with a boolean flag (e.g. enable), so if it is `true` it enables the feature, otherwise, it disables the feature. If my suggestion looks fine, we can go further, using same proposal for `ProposalAddToValidatorWhitelist` and `ProposalRemoveFromValidatorWhitelist`. We can name it `ProposalModifyValidatorWhitelist` and add operation enum which contains `add` and `remove`.

It's like a matter of choice. If the proposal type is various, one option can be reduced, and if the type is reduced, an option will be added. I think you can choose one that is easier to implement.

PR #406, opened 2021-12-17
feat: Add CreateValidator access control feature

0Tech:
## Description

The first requirement of the consortium is an access control of `CreateValidator` message. To implement this feature while not modifying the existing modules, this patch introduces two new modules.

`consortium` module, the first one, has two proposals. `UpdateConsortiumParamsProposal` changes the parameter of the module by which you may disable the entire consortium related features. You CANNOT enable the module again after disabling it, because it would bring the public chain back to the consortium chain. Therefore, you must enable it by changing the default genesis state (the default is off). The second one, `UpdateValidatorAuthsProposal` changes the authorization information of the validators. It is incremental now, but the semantic is subject to change (it could be restful-like). The module has no transactions, but has three queries, one for querying the parameters, and the others for querying validator authorizations.

`stakingplus` module, which was not included in the first plan, was introduced to mitigate a possible DoS attack vector. Basically, it is a thin wrapper of the original `staking` module, and it queries validator auth infos from the consortium module. If an operator attempts to create a validator without being authorized first, the transaction will be rejected.

closes: #384 

---

write a little note why.


x/consortium/types/proposal.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (p *UpdateConsortiumParamsProposal) GetTitle() string       { return p.Title }
+func (p *UpdateConsortiumParamsProposal) GetDescription() string { return p.Description }
+func (p *UpdateConsortiumParamsProposal) ProposalRoute() string  { return RouterKey }
+func (p *UpdateConsortiumParamsProposal) ProposalType() string   { return ProposalTypeUpdateConsortiumParams }
+func (p *UpdateConsortiumParamsProposal) ValidateBasic() error   {
+	params := p.Params
+	if params.Enabled {
+		return ErrInvalidParams
```
0Tech, line 37: Replace the custom error with an sdkerror (e.g. ErrInvalidRequest).

x/consortium/types/keys.go
```diff
@@ -0,0 +1,64 @@
...
+// - 0x00: Params
+//
+// - 0x01<valAddress_Bytes>: bool
+//
+// - 0x02<valAddress_Bytes>: bool
+var (
+	ParamsKey = []byte{0x00}
+	ValidatorAuthKeyPrefix = []byte{0x01}
+
+	PendingRejectedDelegationKeyPrefix = []byte{0x10}
```
0Tech, line 36: Remove the state.

x/consortium/types/genesis.go
```diff
@@ -0,0 +1,38 @@
...
+		ValidatorAuths: []*ValidatorAuth{},
+	}
+}
+
+// ValidateGenesis validates the provided genesis state to ensure the
+// expected invariants holds.
+func ValidateGenesis(data GenesisState) error {
+	// validator auths are redundant where consortium is off
+	if !data.Params.Enabled && len(data.ValidatorAuths) != 0 {
+		return ErrInvalidParams
```
0Tech, line 28: Replace the error with an sdkerror.

x/consortium/types/expected_keepers.go
```diff
@@ -0,0 +1,19 @@
...
+	stakingtypes "github.com/line/lbm-sdk/x/staking/types"
+)
+
+type (
+	StakingKeeper interface {
+		GetDelegation(ctx sdk.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress) (delegations stakingtypes.Delegation, found bool)
+		Undelegate(ctx sdk.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress, sharesAmount sdk.Dec) (time.Time, error)
+		IterateValidators(ctx sdk.Context, fn func(index int64, validator stakingtypes.ValidatorI) (stop bool))
+		// jailValidator(ctx sdk.Context, validator stakingtypes.ValidatorI)
+		GetValidator(ctx sdk.Context, addr sdk.ValAddress) (validator stakingtypes.Validator, found bool)
```
0Tech, line 16: Remove unused methods from the interface.

x/consortium/types/errors.go
```diff
@@ -0,0 +1,12 @@
+package types
```
0Tech, line 1: Remove custom errors.

x/consortium/keeper/validator.go
```diff
@@ -0,0 +1,124 @@
...
+	defer iter.Close()
+	for ; iter.Valid(); iter.Next() {
+		delAddr, valAddr := types.SplitPendingRejectedDelegationKey(iter.Key())
+		if cb(delAddr, valAddr) {
+			break
+		}
+	}
+}
+
+func (k Keeper) RejectDelegations(ctx sdk.Context) {
```
0Tech, line 83: Remove it

egonspace (approved):
Looks great!

dudong2 (approved):
Great!

zemyblue:
The `proposal_handler` does not have event log. 
I think It's good to add the handling result as event log.

x/consortium/client/cli/tx.go
```diff
@@ -0,0 +1,211 @@
...
+						OperatorAddress: addr,
+						CreationAllowed: false,
+					}
+					auths = append(auths, auth)
+				}
+
+				return auths, nil
+			}
+
+			auths, err := createAuths(addingValidators, deletingValidators)
```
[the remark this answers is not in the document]
0Tech: I was concerned about the case too, so it checks the duplication in the function and if it's the case, it will emit an error.

0Tech:
> The `proposal_handler` does not have event log. I think It's good to add the handling result as event log.

I added events to the module. Thank you for your suggestion.

x/consortium/client/cli/tx.go
```diff
@@ -0,0 +1,211 @@
+package cli
```
zemyblue, line 1: I didn't find unittest of this cli codes. 
How about adding unittest of thease cli commands?

You can reference how to developing the cli unittest in https://github.com/line/lbm-sdk/blob/main/x/bank/client/cli/cli_test.go

x/consortium/client/cli/tx.go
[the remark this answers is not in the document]
zemyblue: And please add a unittest for this case.

x/consortium/client/cli/tx.go
0Tech: I will make a new issue on this.
If you agree, please approve this PR :)

x/consortium/client/cli/tx.go
```diff
@@ -0,0 +1,211 @@
...
+					}
+					auths = append(auths, auth)
+				}
+				for _, addr := range deletings {
+					auth := &types.ValidatorAuth{
+						OperatorAddress: addr,
+						CreationAllowed: false,
+					}
+					auths = append(auths, auth)
+				}
```
dudong2, line 179: Why don't ValidatorAuth clear the address and just set CreationAllowed to false?

0Tech: May I ask you more detailed explanation on it?

dudong2: The CreationAllowed value seems to indicate if that OperatorAddress can be created as a validator.
However, I think it should be whether the address is registered as a key in the store, not whether it is determined through a boolean variable.
In other words, the CreationAllowed variable is unnecessary, and I think that the criterion for judging whether the validator can be created is based on whether the address is registered as a key in the store.

0Tech: It's very good question. As you can see, I'm also thinking about replacing that structure into a boolean value or just place holder (the first comment in my self-review). If consortium module has few features, then I will change the states. However, if the requirements involve many authorization flags on validators (e.g. UnbondAllowed, DelegationAllowed, ...), then to simplify queries and setters, it will remain.
This PR is not final, there are one or two refactorings on my schedule.

dudong2: Thanks for the great answer. 👍

zemyblue:
If you change some profo files. Please check the `client/docs/statik/statik.go` file is changed.
When I check it, the proto was changed, but the `statik.go` file was not changed. There may be no changes to the `statik.go` file, but I have no idea what was modified in `statik.go`. So please check it.

zemyblue (approved):

Issue #473, opened 2022-03-22
default output method was changed to 'text' from 'json'


brew0722:
## Summary of Bug

As shown below, in the past, the default output method was json.
`  -o, --output string            Output format (text|json) (default "json")`
The cli still tells me that the default is json.

However, the result output is text by default now.

The defaults below appear to have been changed.
https://github.com/line/lbm-sdk/blob/393d1fcd3f3aaf2c632eff74d21d380d70e60a96/client/config/config.go#L15

## Version

lbm v0.3.0-rc0
lbm-sdk v0.45.0-rc0

## Steps to Reproduce

____

## For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

0Tech:
If you don't mind, can you describe your expectation more specifically?

e.g.)
On queries: text
On transactions: json

The example above is the result produced on my local machine.
And I assumed that there is a typo on the version you specified, so I tested it on version 0.43.0-rc0.

brew0722:
@0Tech 
On queries: json
On transactions: json

I'll add an example.
As far as I know, this is common to all response outputs of lbm.


Below is the default output.
> user@AL01696733 sample-token % lbm tx wasm store target/wasm32-unknown-unknown/release/sample_token.wasm  --from alice --keyring-backend test --chain-id lbm --gas 1000000 -b block -y
> 
> code: 0
> codespace: ""
> data: 0A100A0A73746F72652D636F646512020809
> events:
> - attributes:
>   - index: true
>     key: YWNjX3NlcQ==
>     value: bGluazF0d3NmbXVqMjhuZHBoNTRrNG53OGNyd3U4aDljOG1oM3J0eDcwNS84
>   type: tx
> - attributes:
>   - index: true
>     key: c2lnbmF0dXJl
>     value: NUpJSmxiUkpCV1FoK3FhcXlCd2Z0Ly9rQjZNS09wNWRXcWZhRFdzTzZuMXN3MWx0eG5TZDlYR2l5OWJRdjBmVk9obFJBOGtSV0hReTA2Q21mT3JZUXc9PQ==
>   type: tx
> - attributes:
>   - index: true
>     key: YWN0aW9u
>     value: c3RvcmUtY29kZQ==
>   type: message
> - attributes:
>   - index: true
>     key: bW9kdWxl
>     value: d2FzbQ==
>   - index: true
[...]

PR #476, opened 2022-03-24
fix: change the default value of the client output format in the config

0Tech:
## Description

It fixes the confusing behavior discussed in #473.
It's not a bug, but we concluded not to set the default value on the output format in the config file.

closes: #473

## Motivation and context

#473 

## How has this been tested?

Executed tx cli and confirmed the output is in json.
Also confirmed the output of query cli is in yaml.

Tested on Docker image, on Debian bookworm.

## Checklist:


brew0722 (approved):
LGTM

Issue #482, opened 2022-03-29
Update documents on x/stakingplus


0Tech:
## Summary
Update documents on x/stakingplus which extends x/staking (and introduces breaking changes).

## Problem Definition
It is a part of the corresponding issue #481.

## Proposal
Add all the spec files as x/staking did. Refer the original documents in the files, and if there is any difference, describe it.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #483, opened 2022-03-29
docs: update documents on x/stakingplus

0Tech:
## Description

closes: #482 

## Motivation and context

Refer to #482.

## Checklist:

- [ ] I have added tests to cover my changes. [no need]
- [ ] I have updated the documentation accordingly. [no need]
- [ ] I have updated API documentation `client/docs/swagger-ui/swagger.yaml` [no need]

### DIFF
--- a/x/stakingplus/spec/01_state.md
+++ b/x/stakingplus/spec/01_state.md
@@ -4,4 +4,4 @@ order: 1
 
 # State
 
-There is no difference in state with that of Staking module of the Cosmos-SDK. Refer to the [original document](../../staking/01_state.md) for more information.
+There is no difference in state with that of Staking module of the Cosmos-SDK. Refer to the [original document](../../staking/spec/01_state.md) for more information.
--- a/x/stakingplus/spec/01_state.md
+++ b/x/stakingplus/spec/01_state.md
@@ -1,0 +1,7 @@
+<!--
+order: 1
+-->
+
+# State
+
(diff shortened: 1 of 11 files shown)


dudong2 (approved):

x/stakingplus/spec/01_state.md
```diff
@@ -0,0 +1,7 @@
+<!--
+order: 1
+-->
+
+# State
+
+There is no difference in state with that of Staking module of the Cosmos-SDK. Refer to the [original document](../../staking/spec/01_state.md) for more information.
```
zemyblue, line 7: How about using `LBM-SDK` instead of `Cosmos-SDK`?

0Tech: Do we have to change all the Cosmos things in the specs? I think it's also a good option to describe only the differences between the original documents. Then, the original documents could be referenced by the permanent links, including the version.

If we are planning to change the x/staking spec files in #488, it's plausible to apply your suggestion. However, it appears to be quite opposite to the existence of this module, because all the modifications should be in x/stakingplus, so there would be no differences between x/staking and that of Cosmos-SDK. Then, the documents on x/staking would remain intact. OR, we could write documents on x/staking, including the statements saying there is no difference between the module and the upstream's, and the reference to the original Cosmos-SDK's document. In this case, changing the name to LBM-SDK should be made. But, why should we?

zemyblue: OK, I understood. I think it's better not fix it. Thank you.

zemyblue (approved):

Issue #506, opened 2022-04-11
add foundation treasury feature to x/foundation


0Tech:
## Summary

add foundation treasury feature to x/foundation.

## Proposal

1. The chain collects another tax which is the main source of the foundation treasury.
    - x/foundation sends certain amount of coins from the "fee-collector" to its module account "treasury" BEFORE the x/distribution touches it.
    - The default percentage would be 0%. You may change the percentage in the genesis.
2. Only the authorized operator can withdraw coins from the module account.

TODO: add items after the confirmation.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

PR #518, opened 2022-04-28
feat: add foundation treasury feature to x/foundation

0Tech:
## Description

closes: #506

## Motivation and context

Refer to #506.

## How has this been tested?

Added corresponding unit tests in x/foundation.

## Checklist:


proto/lbm/foundation/v1/foundation.proto
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// Member represents a foundation member with an account address,
+// non-zero weight and metadata.
+message Member {
+  // address is the member's account address.
+  string address = 1;
+
+  // weight is the member's voting weight that should be greater than 0.
+  string weight = 2 [(gogoproto.nullable) = false, (gogoproto.customtype) = "github.com/line/lbm-sdk/types.Dec"];
```
zemyblue, line 55: Why does this weight need? I heard all member has same voting power.

x/foundation/keeper/abci.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// BeginBlocker withdraws rewards from fee-collector before the distribution
+// module's withdraw.
+func BeginBlocker(ctx sdk.Context, k Keeper) {
+	defer telemetry.ModuleMeasureSince(foundation.ModuleName, time.Now(), telemetry.MetricKeyBeginBlocker)
+
+	if err := k.CollectFoundationTax(ctx); err != nil {
+		panic(err)
+	}
```
zemyblue, line 18: We discussed about this reward ratio, and I have known the treasury reward also distributed the `distribution` module for equaling between the each ratio and genesis ratio. Isn't it? But this code it is not. 

This rule was changed?

x/foundation/foundation.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+		addrs[proposer] = true
+
+		if err := sdk.ValidateAccAddress(proposer); err != nil {
+			return sdkerrors.ErrInvalidAddress.Wrapf("invalid proposer address: %s", proposer)
+		}
+	}
+
+	return nil
+}
```
zemyblue, line 52: I cannot found the codes that verify the signatures of multi proposers.
Where is it?

x/foundation/keeper/exec.go
```diff
@@ -0,0 +1,91 @@
...
+		if err := k.doTallyAndUpdate(ctx, proposal); err != nil {
+			return err
+		}
+	}
+
+	// Execute proposal payload.
+	if proposal.Status == foundation.PROPOSAL_STATUS_CLOSED &&
+		proposal.Result == foundation.PROPOSAL_RESULT_ACCEPTED &&
+		proposal.ExecutorResult != foundation.PROPOSAL_EXECUTOR_RESULT_SUCCESS {
+		logger := ctx.Logger().With("module", fmt.Sprintf("x/%s", foundation.ModuleName))
```
zemyblue, line 45: This `logger` is only called when the error of `doExecuteMsgs` is occurred. So How about changing the location to inner of error of `doExecuteMsgs`.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 ### Improvements
 
 * (refactor) [\#493](https://github.com/line/lbm-sdk/pull/493) restructure x/consortium
+* (x/foundation) [\#518](https://github.com/line/lbm-sdk/pull/518) add foundation treasury feature to x/foundation
```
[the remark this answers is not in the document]
0Tech: I moved the entry.

x/foundation/keeper/msg_server.go
```diff
@@ -0,0 +1,268 @@
...
+
+import (
+	"context"
+
+	sdk "github.com/line/lbm-sdk/types"
+	sdkerrors "github.com/line/lbm-sdk/types/errors"
+	"github.com/line/lbm-sdk/x/foundation"
+)
+
+const gasCostPerIteration = uint64(20)
```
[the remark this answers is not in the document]
0Tech: I borrowed the value from x/group of the cosmos-sdk.

x/foundation/foundation.go
0Tech: The module does not verify the signatures, but the antehandler does.

x/foundation/keeper/exec.go
0Tech: The logger is here for the logs to be left in the original context.

proto/lbm/foundation/v1/foundation.proto
```diff
@@ -0,0 +1,280 @@
...
+      [(gogoproto.customtype) = "github.com/line/lbm-sdk/types.Dec", (gogoproto.nullable) = false];
+}
+
+// ValidatorAuth defines authorization info of a validator.
+message ValidatorAuth {
+  string operator_address = 1;
+  bool   creation_allowed = 2;
+}
+
+// UpdateFoundationParamsProposal details a proposal to update params of cosortium module.
```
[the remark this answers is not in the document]
0Tech: I also found I haven't moved the original spec from the old path.

proto/lbm/foundation/v1/tx.proto
```diff
@@ -0,0 +1,182 @@
...
+  repeated lbm.base.v1.Coin amount = 2
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/line/lbm-sdk/types.Coins"];
+}
+
+// MsgFundTreasuryResponse defines the Msg/FundTreasury response type.
+message MsgFundTreasuryResponse {}
+
+// MsgWithdrawFromTreasury represents a message to withdraw coins from the treasury.
+message MsgWithdrawFromTreasury {
+  string   operator                   = 1;
```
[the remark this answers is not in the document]
0Tech: I applied the linter.

proto/lbm/foundation/v1/tx.proto
```diff
@@ -0,0 +1,182 @@
...
+
+// MsgUpdateMembers is the Msg/UpdateMembers request type.
+message MsgUpdateMembers {
+  // operator is the account address of the foundation operator.
+  string operator = 1;
+
+  // member_updates is the list of members to update,
+  // set weight to 0 to remove a member.
+  repeated Member member_updates = 2 [(gogoproto.nullable) = false];
+}
```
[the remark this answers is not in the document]
0Tech: 0 of `weight` means the member has no weight, i.e. one has no voting power. It virtually means he or she is not a member. I think it makes sense.
Removing weight from `Member` leads one to send two messages, one adding members, the other removing some other members. In the current design, `MsgUpdateMembers` allows one to remove and add members in a one message.
So, if we are not planning to force one to send two messages, the only change left is replacing the type from `Dec` to `bool`, and changing the corresponding field name. We may try it, but I think we must have some strong reason to do that.

proto/lbm/foundation/v1/foundation.proto
0Tech: The planning team decided to keep the current design. So I forced the value to be 0 or 1, which is included in the corresponding `ValidateBasic()`.

x/foundation/keeper/abci.go
0Tech: The planning team decided to keep the current design.
If the decision would be changed, I will introduce another wrapper module for x/distribution. In that case, changing the tax rate would be a parameter of x/distribution, which means we need to get a confirm from x/gov to change the value.

proto/lbm/foundation/v1/tx.proto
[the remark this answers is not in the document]
zemyblue: I wonder if the members has no weight are removed from the db. If then, I don't understand why the weight is necessary.

x/foundation/keeper/msg_server.go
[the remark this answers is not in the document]
zemyblue: Is this value appropriate? Please check it later.

x/foundation/keeper/treasury.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func (k Keeper) CollectFoundationTax(ctx sdk.Context) error {
+	// fetch and clear the collected fees for the fund, since this is
+	// called in BeginBlock, collected fees will be from the previous block
+	feeCollector := k.authKeeper.GetModuleAccount(ctx, k.feeCollectorName)
+	feesCollectedInt := k.bankKeeper.GetAllBalances(ctx, feeCollector.GetAddress())
+	feesCollected := sdk.NewDecCoinsFromCoins(feesCollectedInt...)
+
+	// caculate the tax
```
zemyblue, line 15: `calculate` typo

x/foundation/keeper/abci.go
zemyblue: The consensus of validators is also required to modify the value of `FoundationTax`. So it is equals. 
But It seems that it is not fair in the blockchain system to first deduct the total reward value and then divide the remainder between the validator and the proposer.

proto/lbm/foundation/v1/tx.proto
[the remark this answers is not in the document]
0Tech: According to x/group, one may use `MsgUpdateGroupMember` to update the group members. To remove a member, you must set `weight` to 0.
In the current design of x/foundation, `MsgUpdateMember` corresponds to `MsgUpdateGroupMember`. Therefore, if you set the member's `weight` to 0 in `MsgUpdateMember`, the member would be removed.

We cannot just remove the field `weight` from `Member`, for the reason I stated:
> So, if we are not planning to force one to send two messages, the only option left is replacing the type from `Dec` to `bool`, and changing the corresponding field name.

x/foundation/keeper/msg_server.go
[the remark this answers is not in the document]
0Tech: We have to revisit this once later. As of now, the upstream is also not ready.

x/foundation/keeper/abci.go
0Tech: I'm not against to modifying the value by x/gov proposal.
I stated the plan because the cost required to redesign foundation tax rate is rather high. And that was the main reason for the planning team to decide to keep the current design.

zemyblue: If then, how about adding this as another issue and implementing it next time?

x/foundation/keeper/msg_server.go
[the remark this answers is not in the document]
zemyblue: OK, I understood. Let's check it after merging this PR. 
Thank you.

proto/lbm/foundation/v1/tx.proto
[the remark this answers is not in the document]
zemyblue: Ok, I also think it is necessary to replace the type of `weight` to `bool`. 
Thank you.

proto/lbm/foundation/v1/foundation.proto
```diff
@@ -0,0 +1,281 @@
...
+  option (gogoproto.goproto_getters)  = false;
+  option (gogoproto.goproto_stringer) = false;
+
+  string                 title       = 1;
+  string                 description = 2;
+  repeated ValidatorAuth auths       = 3 [(gogoproto.nullable) = false];
+}
+
+// Member represents a foundation member with an account address,
+// non-zero weight and metadata.
```
zemyblue, line 49: Please check  the description about `non-zero` is ok.

proto/lbm/foundation/v1/foundation.proto
```diff
@@ -0,0 +1,281 @@
...
+  // operator is the account address of the foundation's operator.
+  string operator = 1;
+
+  // version is used to track changes to the foundation's membership structure that
+  // would break existing proposals. Whenever any members weight is changed,
+  // or any member is added or removed this version is incremented and will
+  // cause proposals based on older versions of the foundation to fail
+  uint64 version = 2;
+
+  // total_weight is the sum of the foundation members' weights.
```
zemyblue, line 135: also check please

x/foundation/keeper/keys.go
```diff
@@ -0,0 +1,113 @@
...
+func validatorAuthKey(valAddr sdk.ValAddress) []byte {
+	key := make([]byte, len(validatorAuthKeyPrefix)+len(valAddr))
+	copy(key, validatorAuthKeyPrefix)
+	copy(key[len(validatorAuthKeyPrefix):], valAddr)
+	return key
+}
+
+// memberKey key for a specific member from the store
+func memberKey(address sdk.AccAddress) []byte {
+	key := make([]byte, len(memberKeyPrefix)+len(address))
```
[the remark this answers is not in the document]
0Tech: We don't need to store the length of the address which comes last, because we can derive it by subtracting the length of the previous bytes from the entire length.

proto/lbm/foundation/v1/foundation.proto
0Tech: Updated statements related to `weight`.

x/foundation/keeper/keys.go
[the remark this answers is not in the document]
dudong2: I wrote a review for the unity of the key storage method, but when I think about it, the gas will increase as the length of the key gets longer... Then I think your method is better :)

zemyblue (approved):

dudong2 (approved):

Issue #412, opened 2022-01-04
Migrate `token` and `collection` module from link


0Tech:
## Summary

Migrate token and collection module from line/link.

____

#### For Admin Use

- [x] Not duplicate issue
- [x] Appropriate labels applied
- [x] Appropriate contributors tagged
- [x] Contributor assigned/self-assigned

zemyblue:
`Collection` module not ready.

PR #539, opened 2022-05-12
fix: fix the compatibility issues with daphne

0Tech:
## Description

ref: #412

## Motivation and context

The module was not compatible with the legacy one, and we concluded to maintain the most part of the interface.

## Checklist:


proto/lbm/token/v1/tx.proto
```diff
@@ -11,145 +11,172 @@ option (gogoproto.goproto_getters_all) = false;
...
-message MsgTransferFrom {
-  string class_id = 1;
-  string proxy    = 2;
-  string from     = 3;
-  string to       = 4;
-  string amount   = 5 [(gogoproto.customtype) = "github.com/line/lbm-sdk/types.Int", (gogoproto.nullable) = false];
+// MsgOperatorSend defines the Msg/OperatorSend request type.
+message MsgOperatorSend {
+  string contract_id = 1;
+  string proxy       = 2;
```
ryu1-sakai, line 63: The message name is using the work "operator", but the field name is using the word "proxy".
It's better to unify the word.

0Tech: Because I cannot change the field name, I think I should preserve `MsgTransferFrom`. The term `operator` would be used only in the new interface (new queries, new transactions, and new events).

ryu1-sakai (commented):
@0Tech, I wrote one comment.

proto/lbm/token/v1/query.proto
```diff
@@ -11,125 +11,237 @@ option go_package = "github.com/line/lbm-sdk/x/token";
...
+  string holder = 3;
 }
 
-// QueryApproveResponse is the response type for the Query/Approve RPC method
-message QueryApproveResponse {
-  Approve approve = 1;
+// QueryAuthorizationResponse is the response type for the Query/Authorization RPC method
+// Since: finschia
+message QueryAuthorizationResponse {
+  // if no authorization found for the request, it would be nil and return an error.
```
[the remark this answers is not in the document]
0Tech: > Could you describe what kind of error will be sent to clients?

The error would be `types/errors.ErrNotFound`.

> Such information is necessary for client developers to implement their clients.

I will consider documenting it on x/token and x/collection.

> And I wonder whether the error response can contain `QueryAuthorizationResponse`.
> You wrote "it would be nil and return an error", so it seems that the error response has the field `authorization`. Right?

My fault. The entire design of the responses should be fixed, because as you pointed out, on the error, the response could be nil and I don't need to fill out the field. So I will set the `nullable` option of the field to `false`, which means the field must have a sane value in every occasion (where the error is `nil`).

proto/lbm/token/v1/event.proto
```diff
@@ -3,75 +3,168 @@ package lbm.token.v1;
...
+// EventModified is emitted when the information of a token class is modified.
+//
+// Since: 0.46.0 (finschia)
+message EventModified {
+  // contract id associated with the token class.
+  string contract_id = 1;
+  // address which triggered the modify.
+  string operator = 2;
+  // changes on the metadata of the class.
+  repeated Pair changes = 3 [(gogoproto.nullable) = false];
```
ryu1-sakai, line 169: How about emitting multiple events rather than one event with multiple changes?

For example, let's think about a case where `name` and `symbol` are modified.

(1) In case of your idea, the following one event will be emitted.

```
{
  "contract_id": "<contract ID>",
  "operator": "<operator>",
  "changes": [
    { "key": "name", "value": "<new name>" },
    { "key": "symbol", "value": "<new symbol>" }
  ]
}
```

(2) In another idea, the following two events will be emitted.

```
{
  "contract_id": "<contract ID>",
  "operator": "<operator>",
  "field": "name",
  "value": "<new name>"
}

{
  "contract_id": "<contract ID>",
  "operator": "<operator>",
  "field": "symbol",
  "value": "<new symbol>"
}
```

**The advantage of (2) is, clients do not need to parse JSON** to interpret modify events.
In case of (1), clients will need to parse JSON because, as you said before, the field `changes` will be a JSON string in transactions which clients receive via Mainnet RPC.

ryu1-sakai (commented):
@0Tech, I wrote two questions and one comment about `event.proto`.

proto/lbm/token/v1/event.proto
0Tech: > **The advantage of (2) is, clients do not need to parse JSON** to interpret modify events.

I found there is already an implementation of converting events to typed events. The client or the developer of sdk in another language may refer to this code:

https://github.com/line/lbm-sdk/blob/5eee50ecc2c08acbc02935b50d3b69e8ce3654eb/types/events.go#L104-L139

(It's not obvious whether it would be easy to port this code into another language, because it uses some packages)

With that function provided, one could easily restore the events to the corresponding typed events, like:

https://github.com/line/lbm-sdk/blob/5eee50ecc2c08acbc02935b50d3b69e8ce3654eb/types/events_test.go#L74-L103

With that aside, it will be a matter of taste to change the type of the message `EventModify`. If we decide to change the field to the singular one, I think we may also want to change the type of the field `changes` in the tx msg `MsgModify` to the singluar one too (which involves introducing a new tx endpoint and the corresponding msg).

proto/lbm/token/v1/query.proto
```diff
@@ -11,125 +11,237 @@ option go_package = "github.com/line/lbm-sdk/x/token";
...
+  }
+
+  // GranteeGrants queries permissions on a given grantee.
+  // Since: finschia
+  rpc GranteeGrants(QueryGranteeGrantsRequest) returns (QueryGranteeGrantsResponse) {
+    option (google.api.http).get = "/lbm/token/v1/token_classes/{contract_id}/grants/{grantee}";
+  }
+
+  // Authorization queries authorization on a given operator holder pair.
+  // if no authorization found for the request, it would return an error.
```
ryu1-sakai, line 60: Could you write the comment to explain what kind of error clients will receive when no authorization is found?
Client need to distinguish such a not-found error from other kind of errors.

0Tech: I updated the comment.

x/token/keeper/msg_server_test.go
```diff
@@ -5,33 +5,85 @@ import (
 	"github.com/line/lbm-sdk/x/token"
 )
 
-func (s *KeeperTestSuite) TestMsgTransfer() {
+func (s *KeeperTestSuite) TestMsgSend() {
 	testCases := map[string]struct {
-		classId string
+		contractID string
```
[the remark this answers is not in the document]
0Tech: I will apply the lint on the test files, thank you.

dudong2 (approved):
LGTM

PR #580, opened 2022-07-05
fix: push docker image action.

zemyblue:
Signed-off-by: zemyblue <zemyblue@gmail.com>

## Description

The registered docker image of docker hub is not executed in Apple M1 Macbook. Because the docker image was not built with the `platforms` information, so I fix this problem in this PR. 

And `libstdc++` library is required for `linux/amd64` because `bls-eth-go-binary` of `ostracon`. So I add it in the `Dockerfile`.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN make build-linux
 FROM alpine:edge
 
 # Install ca-certificates
-RUN apk add --update ca-certificates
+RUN apk add --update ca-certificates libstdc++
 WORKDIR /root
 
 # Copy over binaries from the build-env
--- a/go.mod
+++ b/go.mod
@@ -62,5 +62,6 @@ require (
 replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	github.com/herumi/bls-eth-go-binary => github.com/herumi/bls-eth-go-binary v0.0.0-20220509081320-2d8ab06de53c
 	google.golang.org/grpc => google.golang.org/grpc v1.33.2
 )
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -30,7 +30,7 @@ RUN make build-linux
 FROM alpine:edge
 
 # Install ca-certificates
-RUN apk add --update ca-certificates
+RUN apk add --update ca-certificates libstdc++
```
0Tech, line 33: Is this change specific to M1? Then, it would be nice to state it in the comment or description of the PR.

tnasu: I know it is for using  BLS of `line/ostracon`.

go.mod
```diff
@@ -62,5 +62,6 @@ require (
 replace (
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.1.7-0.20210622111912-ef00f8ac3d76
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+	github.com/herumi/bls-eth-go-binary => github.com/herumi/bls-eth-go-binary v0.0.0-20220509081320-2d8ab06de53c
```
tnasu, line 65: How about using `line/ostracon`-v1.0.6 instead of using `replace`?

* https://github.com/line/ostracon/tree/v1.0.6

Dockerfile
ulbqb: I tried it on M1 macbook docker linux/arm64 and linux/amd64 environment, but the latest github.com/herumi/bls-eth-go-binary seems to require libstdc++.

go.mod
ulbqb: > How about using line/ostracon-v1.0.6 instead of using replace?

I think so, too.

zemyblue: Thank you. I'll change `line/ostracon` version from `v1.0.6-0.20220614053335-f9d9fa2cc779` to `v1.0.6`. However this change would not have been necessary in the original. :)

tnasu (approved):
LGTM

0Tech (approved):
LGTM

Issue #549, opened 2022-05-30
Reduce the difference between lbm-sdk and cosmos-sdk


zemyblue:
## Summary

Compare the difference between lbm-sdk and cosmos-sdk and reflect the better fix.

## Proposal

- [x] change gRPC path from `lbm-sdk` to `cosmos-sdk`, and from `v1` to `v1beta1`.
- [x] check the performance of `BaseAccount`'s seperated pub_keys and then compare and select better thing. https://github.com/line/lbm-sdk/blob/ac918acc77d77d934e4dff134dcc53b9f83e3370/proto/lbm/auth/v1/auth.proto#L24-L31
- [x] find the difference between cosmos-sdk and lbm-sdk and then make list. (And then apply the better performance logic.)

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

torao:
# List of differences from cosmos-sdk

This comparison is made between the following versions since lbm-sdk 0.46.0-rc*N* are derived from cosmos-sdk 0.45.1.

| Repository | Version | Hash |
| --- | --- | --- |
| line/lbm-sdk | v0.46.0-rc2 | `ac918ac` |
| cosmos/cosmos-sdk | [v0.45.1](https://github.com/cosmos/cosmos-sdk/commits/v0.45.1) | `2646b47` |

## Differential List by Purpose

The source code file names shown here are samples. To see more, you will need to search the project for relevant keywords and explore the reference relationships.

1.  **Performance**
    1.  **Asynchnonization**
        1.  changed: `CheckTx()` → `CheckTxAsync()`, `CheckTxSync()`
            1.  a lot of `Mutex`, `mtx.RLock()`, `mtx.RUnlock()`, `WorkGroup`, `channel` are introduced.
            2.  added: `BeginRecheckTx()`, `EndRecheckTx()`
            3.  mainly `baseapp` (ABCI)
    2.  **Caching**
        2.  added: `store/types/store.go`: caching interface
        3.  added: `store/iavl/cache.go`: caching of IAVL
        4.  added: `store/rootmulti/store.go`: caching of lower-level storage
        5.  added: `baseapp/BaseApp`: `IAVLCacheManager` option
    3.  **Multi-Threading**
        1.  added: `store/cachemulti/store.go` : parallelization of `Write()` process
[...]

PR #588, opened 2022-07-11
Revert BaseAccount back to cosmos-sdk

torao:
## Description

This PR brings the performance fixes mainly done in https://github.com/line/lbm-sdk/commit/39afd48c364044ff6dd45fd8bab7425fb09c8b41 back into the cosmos-sdk implementation based on the results of the performance comparison in #549.

* **PubKey Fields**: Make the `BaseAccount` Protocol Buffers definition have a single field for each type of public-key present.
* **JSON/JSONPB**: Drop the JSON and JSONPB marshaling implementation of `BaseAccount` and revert to using ProtocolBuffers for serialization.
* **Gas Calculation**: Gas consumption is affected since there is a change in the byte size input/output to/from KVS.

closes: #549


x/auth/vesting/types/vesting_account_test.go
```diff
@@ -804,87 +803,21 @@ func TestPermanentLockedAccountMarshal(t *testing.T) {
 	require.NotNil(t, err)
 }
 
-func TestBaseVestingAccountMarshalJSONPB(t *testing.T) {
+func TestBaseVestingAccountMarshal(t *testing.T) {
```
tnasu, line 806: Remove this:
* https://github.com/line/lbm-sdk/pull/517/files#diff-79cdc833ea7de9c9dcb705ff51d3b0d2648ce639a7438c85e5a8cea2671577b1R807

tnasu (commented):
Let's revert too:
* https://github.com/line/lbm-sdk/pull/320/files#diff-085b7c8caa4b62d6ef4bb7a64252bdd927015fe83cbd4f3c62ec58d8741f0d73R106
* https://github.com/line/lbm-sdk/pull/320/files#diff-876463e36c6e6e6ca0eed52e420d4225e3fca76cf84807faba3a8ebf9ba77fcfR68
* etc.

I've checked related PRs:
* https://github.com/line/lbm-sdk/pull/320/files
* https://github.com/line/lbm-sdk/pull/326/files
* https://github.com/line/lbm-sdk/pull/517/files

I think we can more revert. I'll recommend you the reverts after I pick them.

dudong2:
I think the gas consumption would have been reduced because 4 fields were merged into 1. Please check the gas comsumption value in the test codes.

x/auth/vesting/types/vesting_account.go
```diff
@@ -198,54 +196,6 @@ func (bva BaseVestingAccount) MarshalYAML() (interface{}, error) {
 	return marshalYaml(out)
 }
 
-type vestingAccountJSON struct {
```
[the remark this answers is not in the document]
torao: I left a few of these test cases because I thought it would be less likely to cause conflicts if only they were only test cases like this, but well, I'll remove them.

x/auth/vesting/types/vesting_account_test.go
torao: ditto

torao:
@tnasu All formatting differences from cosmos-sdk v0.45.1 in files under `proto/` have been fixed.

@dudong2 Based on preliminary findings, the Marshal Size of the reverted version in one field has increased by 35B (see also https://github.com/line/lbm-sdk/issues/549#issuecomment-1169633561). Perhaps the 4-fields version is stored in binary, whereas the 1-field version is stored in `types.Any` with the relevant information. The gas consumption is calculated at 30 gas per 1 byte and is therefore considered to be increasing.

zemyblue (approved):

tnasu:
@torao I've confirmed `x/bank/` module changes, would you revert them? See the PRs:
* https://github.com/line/lbm-sdk/pull/320/files
* https://github.com/line/lbm-sdk/pull/517/files

>All formatting differences from cosmos-sdk v0.45.1 in files under proto/ have been fixed.

Got it.

torao:
@tnasu I've made the changed files contained in those two PRs identical to the cosmos-sdk as much as possible within the scope of this issue.

tnasu (approved):

PR #589, opened 2022-07-11
fix: fix naming collision in x/token enums

0Tech:
## Description
The patch changes the following enum in the protos:
1. Permission
2. EventType
3. AttributeKey

## Motivation and context

Some of the names in the enums were not unique, which caused compile errors on java.

## Checklist:


proto/lbm/token/v1/event.proto
```diff
@@ -7,40 +7,48 @@ import "lbm/token/v1/token.proto";
 
 option go_package = "github.com/line/lbm-sdk/x/token";
 
+// Deprecated: use typed events.
```
zemyblue, line 10: What does this mean? will These `EventType` enum be deprecated?

0Tech: Yes, I meant to deprecate it. `EventType` enum is for the legacy events.
However, the deprecation no longer affects anything, because we will not update the module.

dudong2 (approved):

dudong2 (approved):

PR #597, opened 2022-07-14
fix: tidy up x/foundation

0Tech:
## Description

This patch addresses the following items:
- Apply lint to the test files
- Fix minor document errors
- Add a test for collecting treasury
- Change error to panic where the error is not caused by user

## Checklist:

### DIFF
--- a/x/foundation/keeper/grant.go
+++ b/x/foundation/keeper/grant.go
@@ -11,9 +11,7 @@ func (k Keeper) Grant(ctx sdk.Context, granter string, grantee sdk.AccAddress, a
 		return sdkerrors.ErrInvalidRequest.Wrapf("authorization for %s already exists", authorization.MsgTypeURL())
 	}
 
-	if err := k.setAuthorization(ctx, granter, grantee, authorization); err != nil {
-		return err
-	}
+	k.setAuthorization(ctx, granter, grantee, authorization)
 
 	any, err := foundation.SetAuthorization(authorization)
 	if err != nil {
@@ -24,7 +22,7 @@ func (k Keeper) Grant(ctx sdk.Context, granter string, grantee sdk.AccAddress, a
 		Grantee:       grantee.String(),
 		Authorization: any,
 	}); err != nil {
-		return err
+		panic(err)
 	}
 
 	return nil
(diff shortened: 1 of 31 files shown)


x/foundation/keeper/grpc_query.go
```diff
@@ -206,16 +206,16 @@ func (s queryServer) Grants(c context.Context, req *foundation.QueryGrantsReques
 		_, err := query.Paginate(grantStore, req.Pagination, func(key []byte, value []byte) error {
 			var authorization foundation.Authorization
 			if err := s.keeper.cdc.UnmarshalInterface(value, &authorization); err != nil {
-				return err
+				panic(err)
```
dudong2, line 209: Why did you change "return err" with "panic(err)"?
Is it because if you process it with return, you have to process each of them outside the function?

0Tech: The error is from `UnmarshalInterface`, which means the state is already corrupted. It's a critical bug and cause by the developer (in this case, me), so there is no reason to return it to the users. It also hides the cause of the error from the user, which is good from the security perspective.

x/foundation/keeper/grant.go
```diff
@@ -57,23 +55,21 @@ func (k Keeper) GetAuthorization(ctx sdk.Context, granter string, grantee sdk.Ac
...
 
-func (k Keeper) setAuthorization(ctx sdk.Context, granter string, grantee sdk.AccAddress, authorization foundation.Authorization) error {
+func (k Keeper) setAuthorization(ctx sdk.Context, granter string, grantee sdk.AccAddress, authorization foundation.Authorization) {
 	store := ctx.KVStore(k.storeKey)
 	key := grantKey(grantee, authorization.MsgTypeURL(), granter)
 
 	bz, err := k.cdc.MarshalInterface(authorization)
 	if err != nil {
-		return err
+		panic(err)
```
zemyblue, line 70: Is this `panic` no problem? Because this `setAutorization` function is call in `Grant` function of `foundation` Msg.

0Tech: No problem. The server won't panic, and it just converts it to the error. You can also see them on many other places (e.g. functions like `MustMarshal()` actually panics inside it).

zemyblue (approved):
LGTM

dudong2 (approved):

PR #599, opened 2022-07-18
fix: fix the order of events

0Tech:
## Description

This patch modifies the order in which events are raised. Also, it adds the unit tests on the events.

## Motivation and context

It should assure the following conditions:
- EventIssue (and the corresponding legacy event) is raised before the relevant events on its contract id.
- The format of the legacy grant event on Msg/Issue is different from that on Msg/Grant.

## Checklist:


x/token/event.go
```diff
@@ -51,7 +51,7 @@ func NewEventIssueToken(e EventIssue, grantee, to sdk.AccAddress, amount sdk.Int
 	new := sdk.NewEvent(eventType)
 	for key, value := range attributes {
 		attribute := sdk.NewAttribute(key.String(), value)
-		new.AppendAttributes(attribute)
+		new = new.AppendAttributes(attribute)
```
zemyblue, line 54: The `new` variable conflicts with the 'builtin' function. 
https://pkg.go.dev/builtin#new

0Tech: I fixed them.
TODO: fix typo in the relevant commit title (name -> new).

zemyblue (approved):

dudong2 (approved):

PR #609, opened 2022-07-24
refactor: remove new APIs on x/token

0Tech:
## Checklist:


proto/lbm/token/v1/tx.proto
```diff
@@ -16,39 +16,22 @@ service Msg {
   // - EventSent
   // - transfer (deprecated, not typed)
   // Throws:
-  // - ErrInsufficientFunds:
+  // - ErrInvalidRequest:
   //   - the balance of `from` does not have enough tokens to spend.
   rpc Send(MsgSend) returns (MsgSendResponse);
```
zemyblue, line 21: Why does not this function change to `Transfer`?

proto/lbm/token/v1/tx.proto
```diff
@@ -16,39 +16,22 @@ service Msg {
   // - EventSent
```
zemyblue, line 16: I knew the `EventSent` event isn't existed in daphne version. Is't it? 
But still it is remained. why?

proto/lbm/token/v1/tx.proto
```diff
@@ -60,8 +43,15 @@ service Msg {
   // Since: 0.46.0 (finschia)
   rpc RevokeOperator(MsgRevokeOperator) returns (MsgRevokeOperatorResponse);
```
zemyblue, line 44: Does `RevokeOperator` keep remaining, not removing? It is new added function.

proto/lbm/token/v1/tx.proto
0Tech: Our clients require to add the new typed events to the old APIs.

proto/lbm/token/v1/tx.proto
0Tech: The planning team added it to the spec, so we cannot remove it. Or, we must contact them to remove it here.

proto/lbm/token/v1/tx.proto
0Tech: We cannot change the name of the endpoint from `Send` to `Transfer`, because it would require another protobuf review of the clients.

dudong2 (approved):

zemyblue (approved):

PR #619, opened 2022-08-02
Modify docker build

shiki-tak:
## Description

Change the Dockerfile to use the downloaded static library
Related: https://github.com/line/lbm/pull/76

## Checklist:


tnasu (approved):

zemyblue (commented):
Please add changelog in `CHANGELOG.md`.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Simple usage with a mounted data directory:
-# > docker build --platform="linux/amd64" -t simapp .
+# > docker build --platform="linux/amd64" -t simapp . --build-arg ARCH=x86_64
```
zemyblue, line 2: `--build-arg ARCH=x86_64` option is requirement? 
If then, how about changing hardcode without input option?

shiki-tak: I would prefer to enter x86_64 as an option rather than hardcode it in the Dockerfile, as it is confusing to change the `ARCH` in the Dockerfile for each architecture.

If you want to run the docker build without the option, how about adding a task to the Makefile?

```
ARCH ?= x86_64 // default value
build-docker: go.sum $(BUILDDIR)/
	docker build -t simapp . --build-arg ARCH=$(ARCH)
```

zemyblue: Your idea is good. 
And how about adding `latest` as the `simapp` tag like below. 
```
          docker build -t simapp:latest . --build-arg ARCH=$(ARCH)
```

zemyblue (approved):

Issue #623, opened 2022-08-02
Detaching the custom wasm proto part of lbm-sdk


zemyblue:
## Summary
Modify `cosmwasm/wasm` proto files in proto directory to be same as `cosmwasm/wasmd`(https://github.com/CosmWasm/wasmd/tree/main/proto/cosmwasm/wasm/v1)
And detaching custom `wasm`'s proto part.

## Problem Definition
We changed `x/wasm`'s proto path from #617 to the same as `cosmwasm/wasmd`'s proto path.
`StoreCodeAndInstantiateContract` tx is not exist in original `cosmwasm/wasmd`'s path, but it exists in wasm proto. https://github.com/line/lbm-sdk/blob/9ba6ea37c656a09697e49f81a62e78326f0455f0/proto/cosmwasm/wasm/v1/tx.proto#L18-L19 
This causes compatibility issues with `cosmwasm/wasmd`.

So we need to detaching lbm-sdk's custom wasm proto part including `StoreCodeAndInstantiateContract`.

## Proposal
1. detaching `StoreCodeAndInstantiateContract` from wasm's tx.proto.

2. replace contract blacklist function 
 - Rollback `ContractInfo` struct of wasm's type.proto to origin.
 - remove `ContractStatus` of `ContractInfo`. 
 - make new `genesis` proto in `lbm/wasm/v1`. this genesis is similar with `wasmd`'s genesis, but add the black list function instead of `ContractStatue` of `ContractInfo` struct.

3. add the proposals to add or remove contract black list.

4. add contract black list query API.
____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #646, opened 2022-08-18
feat: add inactive contract management in wasm.

zemyblue:
## Description

related: #623 

This PR is the function that replacing previous contract blacklist management. 
The previous contract blacklist management function was managed in the contract `status` of wasm `genesisStatus` I
But in order that we added this function in the contract `status`, we needed to customize original genesis proto. 

So I add new lbm's genesis proto and add inactive contract list in lbm's `genesisStatus` without modifying original `genesisStatus`.

Therefore, I add new wrapping structure like below UML.
lbmwasm_main

### Change smart contract blacklist function
* The smart contract blacklist is managed in `inactiveContractAddresses` of wasm `GenesisState`.
* Enable to add and remote to blacklist via `DeactivateContractProposal` and `ActivateContractProposal` proposal.
* Add `DeactivateContractProposal` and `ActivateContractProposal` proposal cli.

**Notice** : `InitGenesis` function of wrapped lbm wasm wasn't implemented, because this need to implement in `wasm/keeper` but if then the import cycle problem is occurred. So, I will modify the addition of 'InitGenesis' and the linkage with the blacklist of the 'bankplus' module to another PR.

## Checklist:


zemyblue:
It is difficult to apply our change of gas config setting with lbm wrapping way. Because the gas config setting of `cosmwasm/wasmd` is private field and It can't change with param or wrap the struct. 

So I'll close this PR and try to change the refactoring way.

PR #685, opened 2022-09-27
refactor: remove x/foundation UpdateValidatorAuthsProposal

0Tech:
## Description
This patch removes `UpdateValidatorAuthsProposal` of `x/foundation` which extends `x/gov` proposal. As a result, granter is no more in use, because `x/foundation` would be the only granter.

## Checklist:


zemyblue (commented):
If this patch is applied, the already launched chain should be reset. Is it right?

0Tech:
> If this patch is applied, the already launched chain should be reset. Is it right?

Of course, yes, because it affects the state.

zemyblue (approved):

dudong2 (approved):

PR #686, opened 2022-09-28
refactor: remove `MinThreshold` and `MinPercentage` from x/foundation config

0Tech:
## Description

Remove the minimum boundary of the policies:
- `MinThreshold`
- `MinPercentage`

## Motivation and context

After this patch is applied, we could out-source the proposal relevant logic to `x/group` later.

## Checklist:

### DIFF
--- a/x/foundation/foundation.go
+++ b/x/foundation/foundation.go
@@ -12,9 +12,9 @@ import (
 	sdkerrors "github.com/line/lbm-sdk/types/errors"
 )
 
-func DefaultDecisionPolicy(config Config) DecisionPolicy {
+func DefaultDecisionPolicy() DecisionPolicy {
 	return &ThresholdDecisionPolicy{
-		Threshold: config.MinThreshold,
+		Threshold: sdk.OneDec(),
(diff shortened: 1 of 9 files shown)


x/foundation/foundation.go
```diff
@@ -286,10 +286,6 @@ func (p ThresholdDecisionPolicy) ValidateBasic() error {
 }
 
 func (p ThresholdDecisionPolicy) Validate(config Config) error {
-	if p.Threshold.LT(config.MinThreshold) {
-		return sdkerrors.ErrInvalidRequest.Wrap("threshold must be greater than or equal to min_threshold")
-	}
```
zemyblue, line 291: To remove threadhold checking with Minthreadhold is OK. But I think `Threadold` should be at least greater then 1. Is it right? It then, I think checking is need.

0Tech: The logic is in `ValidateBasic()` of it.

dudong2 (approved):

zemyblue (approved):

PR #698, opened 2022-10-06
refactor: update x/group relevant logic in x/foundation

0Tech:
## Description

This patch updates the proposal related logic of `x/foundation` into the latest `x/group`'s.

The major changes are:
- Split `Member` into `Member` and the new `MemberRequest` which is specific to `Msg/UpdateMembers`.
  - `Member` no more needs `participating` field.
  - New `remove` field is introduced into `MemberRequest`.
- Merge `ProposalResult` into `ProposalStatus`.

## Checklist:


proto/lbm/foundation/v1/foundation.proto
```diff
@@ -34,17 +34,33 @@ message Member {
...
 
   // added_at is a timestamp specifying when a member was added.
   google.protobuf.Timestamp added_at = 4 [(gogoproto.nullable) = false, (gogoproto.stdtime) = true];
 }
 
-// ThresholdDecisionPolicy implements the DecisionPolicy interface
+// MemberRequest represents a foundation member to be used in Msg server requests.
+// Contrary to `Member`, it doesn't have any `added_at` field
+// since this field cannot be set as part of requests.
+message MemberRequest {
```
zemyblue, line 47: This name make confuse for me, there is no query prefix but it looks like a query request. So how about changing `MemberUpdate` or others?

x/foundation/foundation.go
```diff
@@ -156,6 +150,22 @@ func (t TallyResult) TotalCounts() sdk.Dec {
 
 var _ codectypes.UnpackInterfacesMessage = (*Proposal)(nil)
 
+func (p Proposal) ValidateBasic() error {
+	if p.Id == 0 {
+		return sdkerrors.ErrInvalidRequest.Wrap("id must be > 0")
+	}
```
zemyblue, line 156: There is no unittest in this logic.

proto/lbm/foundation/v1/foundation.proto
0Tech: As always, the message is taken from the upstream. Unless there is a critical issue, I think it would be better not to make a change.

zemyblue: This is not a critical issue, but as I mentioned, it make me to confuse. 
BTW, could you share the link of similar upstream message of proto.

x/foundation/foundation.go
0Tech: I added the corresponding unit test.

proto/lbm/foundation/v1/foundation.proto
0Tech: > BTW, could you share the link of similar upstream message of proto.

https://github.com/cosmos/cosmos-sdk/blob/5f01f6f5e51156f8828e0b25e898d801442cd211/proto/cosmos/group/v1/tx.proto#L100
and
https://github.com/cosmos/cosmos-sdk/blob/5f01f6f5e51156f8828e0b25e898d801442cd211/proto/cosmos/group/v1/types.proto#L31-L44

zemyblue (approved):

dudong2 (approved):
LGTM

PR #709, opened 2022-10-12
feat: mint large amount of coins(gov mint)

dudong2:
## Description
It is necessary to mint a large amount of coins. 
For this, x/foundation provides a minting function(one time mint) through proposals.
However, this function is limited to only 1 time.

closes: #XXXX

## Checklist:


x/foundation/keeper/genesis.go
```diff
@@ -114,6 +114,11 @@ func (k Keeper) InitGenesis(ctx sdk.Context, sk foundation.StakingKeeper, data *
 
 	k.SetPool(ctx, data.Pool)
 
+	if data.OneTimeMintMaxCount != 1 {
+		return errors.New("OneTimeMintMaxCount must be 1")
+	}
```
0Tech, line 119: You may remove the validation logic thanks to `ValidateGenesis()`.

x/foundation/keeper/msg_server.go
```diff
@@ -297,3 +297,43 @@ func (s msgServer) Revoke(c context.Context, req *foundation.MsgRevoke) (*founda
...
+	}
+
+	if err := s.keeper.validateOperator(ctx, req.Operator); err != nil {
+		return nil, err
+	}
+
+	if req.Amount.Empty() {
+		return nil, sdkerrors.ErrUnauthorized.Wrapf("The one-time-mint request amount is empty.")
+	}
+
```
0Tech, line 317: It should be moved to `ValidateBasic()` of the corresponding msg.

0Tech (changes requested):
According to the current design, `OneTimeMintMaxCount` has the max count of the mints.
- One executes the minting, then the count is decreased. I think it must not change the value, as the name of value suggests.
- The value is not being exported by `ExportGenesis()`, hence the value in the new genesis would be zero, which is invalid (by `ValidateGenesis()`).
- Even if the value is exported, it may have a zero-value, which will throw an error.
Therefore, I suggest you change the name of the value (e.g. NumberOfMintingLeft), which would mean the number of times you have left, and set the default value to zero. And modify the corresponding codes as the name suggests.

x/foundation/msgs.go
```diff
@@ -353,3 +353,24 @@ func (m MsgRevoke) GetSigners() []sdk.AccAddress {
...
+	if !m.Amount.IsValid() || !m.Amount.IsAllPositive() {
+		return sdkerrors.ErrInvalidCoins.Wrap(m.Amount.String())
+	}
+
+	return nil
+}
+
+// GetSigners implements Msg.
+func (m MsgOneTimeMint) GetSigners() []sdk.AccAddress {
+	signer, _ := sdk.AccAddressFromBech32(m.Operator)
```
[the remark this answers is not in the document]
dudong2: Good suggestion. Thank you!

x/foundation/keeper/genesis.go
dudong2: I removed that logic.

x/foundation/keeper/msg_server.go
dudong2: I moved that codes. thx.

x/foundation/genesis.go
```diff
@@ -10,7 +10,8 @@ import (
 // DefaultGenesisState creates a default GenesisState object
 func DefaultGenesisState() *GenesisState {
 	return &GenesisState{
-		Params: DefaultParams(),
+		Params:               DefaultParams(),
+		OneTimeMintLeftCount: 0,
```
0Tech, line 14: Or, you can omit the field.
```suggestion
		Params: DefaultParams(),
```

x/foundation/msgs.go
```diff
@@ -353,3 +353,28 @@ func (m MsgRevoke) GetSigners() []sdk.AccAddress {
...
+	if _, err := sdk.AccAddressFromBech32(m.Operator); err != nil {
+		return sdkerrors.ErrInvalidAddress.Wrapf("invalid operator address: %s", m.Operator)
+	}
+
+	if m.Amount.Empty() {
+		return sdkerrors.ErrInvalidCoins.Wrapf("The one-time-mint request amount is empty.")
+	}
+
+	if !m.Amount.IsValid() || !m.Amount.IsAllPositive() {
+		return sdkerrors.ErrInvalidCoins.Wrap(m.Amount.String())
```
0Tech, line 370: You may add the unit test for it.

0Tech (commented):
Overall, I think the updated design has no flaws.
One side note: Can you be sure that this "one time" spec won't change? If there is even the slightest chance of such a case, the names of the relevant variables should be changed, and also the literal should be replaced by a config.

zemyblue (commented):
How about using `GovMint` or "governanceMint` the value name instead of `OneTimeMint`

x/foundation/keeper/msg_server.go
```diff
@@ -297,3 +297,39 @@ func (s msgServer) Revoke(c context.Context, req *foundation.MsgRevoke) (*founda
...
+		return nil, err
+	}
+
+	oneTimeMintLeftCount--
+	s.keeper.SetOneTimeMintLeftCount(ctx, oneTimeMintLeftCount)
+
+	if err := ctx.EventManager().EmitTypedEvent(&foundation.EventOneTimeMint{
+		Amount: req.Amount,
+	}); err != nil {
+		panic(err)
```
zemyblue, line 331: When does this panic occur?
Because if the panic occurs, the `LeftCount` decreased but the mint is fail. Isn't it?

proto/lbm/foundation/v1/tx.proto
```diff
@@ -50,6 +50,9 @@ service Msg {
   // Revoke revokes any authorization corresponding to the provided method name on the
   // granter that has been granted to the grantee.
   rpc Revoke(MsgRevoke) returns (MsgRevokeResponse);
+
+  // OneTimeMint defines a one-time mint coins to the treasury.
+  rpc OneTimeMint(MsgOneTimeMint) returns (MsgOneTimeMintResponse);
```
zemyblue, line 55: Please add cli and cli unittest.

zemyblue:
Please update swagger and statik file.

0Tech (changes requested):
I forgot to mention query.proto. You need to add the corresponding endpoint to show the number of minting left.

x/foundation/msgs.go
dudong2: IsValid func is including IsAllPositive() check. So, I deleted IsAllPositive() check and added unit test for that check.

proto/lbm/foundation/v1/tx.proto
dudong2: In the case of one-time mint, it is not appropriate to expose it to the cli because it should be performed only through proposals.

x/foundation/keeper/msg_server.go
dudong2: All other emit events have the same shape. Under normal circumstances, this cannot happen. Also, Is the failure of the event, the consistency in the store such as left count or mint is always maintained.

proto/lbm/foundation/v1/query.proto
```diff
@@ -215,3 +220,12 @@ message QueryGrantsResponse {
...
 }
+
+// QueryGovMintRequest is the Query/GovMint request type.
+message QueryGovMintRequest {}
+
+// QueryGovMintResponse is the Query/GovMint response type.
+message QueryGovMintResponse {
+  // leftCount is the left count of GovMint.
+  uint32 leftCount = 1;
+}
```
0Tech, line 231: Please insert a new line.

x/foundation/genesis.go
```diff
@@ -111,6 +111,10 @@ func ValidateGenesis(data GenesisState) error {
 		return err
 	}
 
+	if data.GovMintLeftCount > 1 {
+		return sdkerrors.ErrInvalidType.Wrap("invalid govMintLeftCount(0 or 1)")
+	}
+
```
0Tech, line 117: The `1` should be replaced by a config, or at least by a constant.
Then, you may correct the corresponding error messages.
Also, I think `ErrInvalidType` here is inappropriate.

x/foundation/keeper/msg_server.go
zemyblue: OK, I understand. Good. Thank you.

proto/lbm/foundation/v1/query.proto
```diff
@@ -69,6 +69,11 @@ service Query {
   rpc Grants(QueryGrantsRequest) returns (QueryGrantsResponse) {
     option (google.api.http).get = "/lbm/foundation/v1/grants/{grantee}/{msg_type_url}";
   }
+
+  // GovMint queries the left count of gov mint.
+  rpc GovMint(QueryGovMintRequest) returns (QueryGovMintResponse) {
+    option (google.api.http).get = "/lbm/foundation/v1/GovMint";
```
0Tech, line 75: Use snake_case.

proto/lbm/foundation/v1/tx.proto
zemyblue: The other proposal tx also has cli and cli test. For example, `WithdrawFromTreasury` and `Grant`. Check it please.

proto/lbm/foundation/v1/query.proto
dudong2: I changed it. Thank you

proto/lbm/foundation/v1/genesis.proto
```diff
@@ -37,6 +37,9 @@ message GenesisState {
 
   // pool
   Pool pool = 8 [(gogoproto.nullable) = false];
+
+  // govMintLeftCount is one time mint max count
```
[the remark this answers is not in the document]
dudong2: I changed all `one time mint` to `gov mint`.

x/foundation/genesis.go
dudong2: I created x/fountdation/errors.go and added appropriate error.

proto/lbm/foundation/v1/tx.proto
dudong2: I added cli and cli test. Thank you.

zemyblue (commented):
Please update `statik.go` file also. And could you change the title more easier to understand about this changes?

zemyblue:
> Please update `statik.go` file also. And could you change the title more easier to understand about this changes?

My system's statik.go is a little difference with upstream. But the code is not difference, just comment and new line. So it is not problem. Sorry.

zemyblue (approved):

0Tech (approved):

PR #712, opened 2022-10-13
fix: fix x/foundation EndBlocker

0Tech:
## Description

This patch will fix `EndBlocker()` of `x/foundation` and add the corresponding unit tests.

## Checklist:

### DIFF
--- a/x/foundation/keeper/keys.go
+++ b/x/foundation/keeper/keys.go
@@ -24,6 +24,9 @@ var (
 	poolKey = []byte{0x30}
 )
 
+// must be constant
+var lenTime = len(sdk.FormatTimeBytes(time.Now()))
+
 // Uint64FromBytes converts a byte array to uint64
 func Uint64FromBytes(bz []byte) uint64 {
 	return binary.BigEndian.Uint64(bz)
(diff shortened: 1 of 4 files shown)


dudong2 (approved):
LGTM

x/foundation/keeper/keys.go
```diff
@@ -70,39 +79,38 @@ func voteKey(proposalID uint64, voter sdk.AccAddress) []byte {
...
-// 	begin := len(proposalByVPEndKeyPrefix) + 1
-// 	end := begin + int(key[begin-1]) // uint64
-// 	proposalID = Uint64FromBytes(key[begin:end])
+func splitProposalByVPEndKey(key []byte) (vpEnd time.Time, id uint64) {
+	prefix := proposalByVPEndKeyPrefix
+	begin := len(prefix)
+	end := begin + lenTime
+	vpEnd, err := sdk.ParseTimeBytes(key[begin:end])
+	if err != nil {
+		panic(err)
```
zemyblue, line 106: Isn't there this panic case?
There is not unittest.

0Tech: You can add a unit test that covers this line. But in the sane condition, it won't happen. There are many similar lines, and of course, x/foundation is not only the case. I think it won't be necessary to add the unit test.

zemyblue: OK, I understand.

zemyblue (approved):

PR #720, opened 2022-10-17
fix: remove potential runtime panic in x/feegrant

ulbqb:
## Description

I added safe type casting and then removed potential runtime panic.

## Motivation and context

Unsafe type casting may cause potential runtime panic.

## Checklist:


x/feegrant/filtered_fee_test.go
```diff
@@ -188,3 +189,80 @@ func TestFilteredFeeValidAllow(t *testing.T) {
...
+func (i invalidAllowance) Accept(ctx sdk.Context, fee sdk.Coins, msgs []sdk.Msg) (remove bool, err error) {
+	return false, nil
+}
+
+func (i invalidAllowance) ValidateBasic() error {
+	return nil
+}
+
+// invalidAllowance2 does not have field
+type invalidAllowance2 struct {
```
0Tech, line 209: You may make it to extend `invalidAllowance1`.

0Tech (approved):
It would be better to provide more meaningful names to the mocks.

x/feegrant/filtered_fee_test.go
tnasu: And I prefer to have specific naming. How about `invalidFieldAllowance`?

0Tech: Is the absence of the field relevant to the failure? It seems like, being `nil` is the trigger, isn't it?

ulbqb: Yes, `nil` is the trigger. I fix the test case name.

x/feegrant/filtered_fee_test.go
```diff
@@ -188,3 +189,73 @@ func TestFilteredFeeValidAllow(t *testing.T) {
...
+		"valid allowance": {
+			allowance: &feegrant.BasicAllowance{},
+			valid:     true,
+		},
+		"invalid interface allowance": {
+			allowance: &invalidInterfaceAllowance{},
+			valid:     false,
+		},
+		"empty allowance": {
+			allowance: (*invalidProtoAllowance)(nil),
```
0Tech, line 241: Do you really need `invalidProtoAllowance`? Is this possible:
```suggestion
			allowance: (*feegrant.BasicAllowance)(nil),
```

ulbqb: That change cause panic at [`proto.Marshal`](https://github.com/line/lbm-sdk/blob/d48d8a7094acb0253e0ba9b279f4ee5de3daef4a/codec/types/any.go#L68).
```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
```

zemyblue (approved):

PR #734, opened 2022-10-18
fix: add restrictions on the number of args in the CLIs

0Tech:
## Description

This patch will add the restrictions on the number of arguments in the CLIs.

## Checklist:


x/bank/client/cli/query.go
```diff
@@ -105,6 +105,7 @@ Example:
 func GetCmdDenomsMetadata() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "denom-metadata",
+		Args:  cobra.NoArgs,
```
loin3, line 108: RunE function takes args as `_` (`RunE: func(cmd *cobra.Command, _ []string)`). 
Is it because `GetCmdDenomsMetadata` does not takes args?

loin3 (commented):
In `x/wasm/client/cli/query.go`, `cobra.ExactArgs(0)` is being used.
Is it same with `cobra.NoArgs`?

x/bank/client/cli/query.go
0Tech: Yes, I think.

loin3: Both of my comments are about inconsistency of code. 
Should these things be fixed in this PR? Or is it better to leave it as an issue in upstream?

0Tech: I think the inconsistency should be resolved by another PR. Because the upstream may reject the modification about it.

loin3 (approved):

x/gov/client/cli/query.go
```diff
@@ -94,6 +94,7 @@ $ %s query gov proposal 1
 func GetCmdQueryProposals() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "proposals",
+		Args:  cobra.NoArgs,
```
zemyblue, line 97: Do you know why this line is not covered?

CHANGELOG.md
```diff
@@ -112,6 +112,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * (crypto) [\#731](https://github.com/line/lbm-sdk/pull/731) remove VRFProve function
 * (x/foundation) [\#732](https://github.com/line/lbm-sdk/pull/732) add verification on accounts into x/foundation Grants cli
 * (x/foundation) [\#730](https://github.com/line/lbm-sdk/pull/730) prune stale x/foundation proposals at voting period end
+* (client) [\#734](https://github.com/line/lbm-sdk/pull/734) add restrictions on the number of args in the CLIs
```
zemyblue, line 115: How about using `cli` as a keyword?

dudong2 (approved):
LGTM

CHANGELOG.md
0Tech:
```suggestion
* (cli) [\#734](https://github.com/line/lbm-sdk/pull/734) add restrictions on the number of args in the CLIs
```

x/gov/client/cli/query.go
0Tech: In the current testing structure for the CLIs, codecov cannot detect that the code is covered. So I cannot increase the coverage right now. But, actually, the tests DOES cover the CLIs.

Issue #717, opened 2022-10-17
Key Could Possibly Be Deleted with `AddKeyCommand()`


tkxkd0159:
## Summary of Bug
The command `AddKeyCommand()` is able to add keys in different modes. In particular, the Multiple Signatures under a key could be added by the `AddKeyCommand()` with the flag `flagMultisig`. The values of the flag `flagMultisig` are the keys in the keyring storage.

The key will be overridden in the keyring by executing `AddKeyCommand()` with the same name. There is an interactive input prompt for user confirmation. Once confirmed, the previous key with the name is deleted from the keyring storage. Then new multiSigKey is saved with the same name.

```go
// client/add.go

	_, err = kb.Key(name)
		if err == nil {
			// account exists, ask for user confirmation
			response, err2 := input.GetConfirmation(fmt.Sprintf("override the existing name %s", name), inBuf, cmd.ErrOrStderr())
			if err2 != nil {
				return err2
			}

			if !response {
				return errors.New("aborted")
			}

			err2 = kb.Delete(name)
			if err2 != nil {
				return err2
			}
		}

		multisigKeys, _ := cmd.Flags().GetStringSlice(flagMultisig)
		if len(multisigKeys) != 0 {
			pks := make([]cryptotypes.PubKey, len(multisigKeys))
			multisigThreshold, _ := cmd.Flags().GetInt(flagMultiSigThreshold)
			if err := validateMultisigThreshold(multisigThreshold, len(multisigKeys)); err != nil {
				return err
			}

			for i, keyname := range multisigKeys {
				k, err := kb.Key(keyname)
				if err != nil {
					return err
				}

				pks[i] = k.GetPubKey()
			}

			if noSort, _ := cmd.Flags().GetBool(flagNoSort); !noSort {
				sort.Slice(pks, func(i, j int) bool {
					return bytes.Compare(pks[i].Address(), pks[j].Address()) < 0
				})
			}

			pk := multisig.NewLegacyAminoPubKey(multisigThreshold, pks)
			info, err := kb.SaveMultisig(name, pk)
			if err != nil {
				return err
			}

			return printCreate(cmd, info, false, "", outputFormat)
		}
```

It is a requirement that the keys exist in the keyring if Multiple Signatures are implemented with the flag `flagMultisig`. If one of the keys is missing in the keyring storage during the execution, then it could lead to an error of the key adding operation by `AddKeyCommand()`, resulting in the deletion of the existing key.

## Version
[...]

PR #737, opened 2022-10-18
fix: check multisig key list to prevent unexpected key deletion

tkxkd0159:
add `verifyMultisigTarget`

## Description

closes: #717 

## How has this been tested?
Test by CLI with the audit scenario.

## Checklist:


client/keys/add.go
```diff
@@ -95,12 +95,13 @@ func runAddCmdPrepare(cmd *cobra.Command, args []string) error {
...
 /*
 input
-	- bip39 mnemonic
-	- bip39 passphrase
-	- bip44 path
-	- local encryption password
+  - bip39 mnemonic
+  - bip39 passphrase
+  - bip44 path
+  - local encryption password
```
loin3, line 101: Is this indent lint done by yourself or intellij?
Cosmos-sdk is same as former one.

loin3 (commented):
This PR does not seem to implement second recommended check `Check if the Multisig contains the same name that will be overridden, and return an error to remind the users of trying a different name.`.

client/keys/add.go
tkxkd0159: > This PR does not seem to implement second recommended check `Check if the Multisig contains the same name that will be overridden, and return an error to remind the users of trying a different name.`.

I didn't create that logic because there is a common flow to check whether to proceed with the override. So the user can decide to override based on his choice.

tkxkd0159: > Is this indent lint done by yourself or intellij? Cosmos-sdk is same as former one.

done by intellij default formatter

loin3: > done by intellij default formatter
When I use the commit feature provided by intellij or goland, lint automatically runs and changes are made in unwanted areas, so I am doing git commit in the terminal without using it. 
As in this comment, if lint runs and make changes in unwanted area, is it better to revert it?

> I didn't create that logic because there is a common flow to check whether to proceed with the override. So the user can decide to override based on his choice.

I thought this recommendation is for the case below.
There is normal keys named `account7, account8` and user want to make new multisig key named `account8` from keys `account7, account8`. In this case, even though user confirm this process, process will be fail as below img.
 
스크린샷 2022-10-19 오후 5 07 16

In upper case, `verifyMultisigTarget`  will return no error and normal key `account8` will be deleted.  But line 160, normal key `account8` is called again which leads an error `key not found`.

So I think that why audit recommends `Check if the Multisig contains the same name that will be overridden, and return an error to remind the users of trying a different name` to prevent creating a multisig key with the same name as the normal key to be used for making multisig key.

loin3: Oh you already added logic to cover my error case at line 350 😄 .

loin3 (approved):

client/keys/add.go
```diff
@@ -327,3 +336,32 @@ func printCreate(cmd *cobra.Command, info keyring.Info, showMnemonic bool, mnemo
...
+	}
+
+	kmap := make(map[string]bool)
+	for i := 0; i < len(kl); i++ {
+		kmap[kl[i].GetName()] = true
+	}
+
+	if _, ok := kmap[newkey]; ok {
+		return errors.New("you cannot specify a new key as one of the names of the keys that make up a multisig")
+	}
```
zemyblue, line 353: How about using `kr.Key(uid string)` or `kr.KeyByAddress(addr sdk.Address)` instead of this all this key traversal?

client/keys/add.go
```diff
@@ -327,3 +336,17 @@ func printCreate(cmd *cobra.Command, info keyring.Info, showMnemonic bool, mnemo
 
 	return nil
 }
+
+func verifyMultisigTarget(kb keyring.Keyring, multisigKeys []string, newkey string) error {
+	if _, err := kb.Key(newkey); err == nil {
```
zemyblue, line 341: In this case, `multisigKeys` should be key name not address. Is it right?

tkxkd0159: yes that's right. It is derived from `flagMultisig`

client/keys/add.go
```diff
@@ -118,6 +120,18 @@ func runAddCmd(ctx client.Context, cmd *cobra.Command, args []string, inBuf *buf
 	if err != nil {
 		return err
 	}
+	multisigKeys, _ := cmd.Flags().GetStringSlice(flagMultisig)
+	if len(multisigKeys) != 0 {
+		multisigThreshold, _ = cmd.Flags().GetInt(flagMultiSigThreshold)
+		if err = validateMultisigThreshold(multisigThreshold, len(multisigKeys)); err != nil {
```
zemyblue, line 126: There is not the unittest of `validateMultisigThreshold` function? Please add it.

zemyblue (approved):

Issue #753, opened 2022-10-25
Add `RequestCheckTxAsync` channel count as config


zemyblue:
## Summary

Need to add the number of `RequestCheckTxAsync` channel as config.

## Problem Definition

In function NewBaseApp(), the size of the channel is fixed to a magic number of 10000. If there could be more than 10000 requests, the concurrent checkTx would be blocked until some goroutine releases in the channel.

https://github.com/line/lbm-sdk/blob/f7d1830f85c2958fc19ae8e7f8695c566fa1a71d/baseapp/baseapp.go#L143-L157

The number of `chCheckTx` channel uses magic nubmer 10000. 

If the size of chCheckTx is larger than the node mempool size, the node mempool will be full of transaction and could handle any more transactions into the pool;

If the size of chCheckTx is smaller than the node mempool size, it could not make full use of the concurrent checkTx feature as designed to improve the overall performance.

So, need to properly setting the `chCheckTx` according to the size of mempool from the underlying node. 

## Proposal

____

#### For Admin Use

- [ ] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #756, opened 2022-10-25
feat: change to create `chCheckTx` with the value set in app config.

zemyblue:
## Description

Change to create `chCheckTx` with the value set in app config.

Basically, `chCheckTx` of BaseApp should be equals to or greater than mempool size. 
But when running NewBaseApp, the mempool size can not know.
So I add the feature set the size of `chCheckTx` as config

So following features have been changed
* add `chCheckTxSize` in Baseapp
* add `FlagChanCheckTxSize` flag as cmd flag
* add `chan-check-tx-size` field in app config.
* set default `chanCheckTxSize` to 10000.

closes: #753 

## Checklist:

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -23,6 +23,9 @@ const (
 
 	// DefaultGRPCWebAddress defines the default address to bind the gRPC-web server to.
 	DefaultGRPCWebAddress = "0.0.0.0:9091"
+
+	// DefaultChanCheckTxSize defines the default maximum size of channel check tx in Baseapp
+	DefaultChanCheckTxSize = 10000
 )
 
 // BaseConfig defines the server's basic configuration
@@ -85,6 +88,9 @@ type BaseConfig struct {
 	// When true, Prometheus metrics are served under /metrics on prometheus_listen_addr in config.toml.
 	// It works when tendermint's prometheus option (config.toml) is set to true.
 	Prometheus bool `mapstructure:"prometheus"`
+
+	// ChanCheckTxSize is the size of RequestCheckTxAsync of BaseApp
+	ChanCheckTxSize uint `mapstructure:"chan-check-tx-size"`
 }
 
 // APIConfig defines the API listener configuration.
@@ -231,6 +237,7 @@ func DefaultConfig() *Config {
 			PruningInterval:     "0",
 			MinRetainBlocks:     0,
 			IndexEvents:         make([]string, 0),
+			ChanCheckTxSize:     DefaultChanCheckTxSize,
 		},
 		Telemetry: telemetry.Config{
 			Enabled:      false,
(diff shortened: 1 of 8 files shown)


dudong2 (approved):

server/config/toml.go
```diff
@@ -86,6 +86,10 @@ index-events = {{ .BaseConfig.IndexEvents }}
 # It works when tendermint's prometheus option (config.toml) is set to true.
 prometheus = {{ .BaseConfig.Prometheus }}
 
+# ChanCheckTxSize is he size of RequestCheckTxAsync of BaseApp
```
tkxkd0159, line 89: 1. typo: ChanCheckTxSize is "the" ~
2. For readability, I think it's necessary to synchronize the comment of `config.go` and `toml.go` for `chan-check-tx-size`.

zemyblue: Thank you for checking the typo. And the comment of `config.go` is generated from Ostracon. So it is out of the scope of this PR. But I'll add the comment to need to check the mempool size of the `config.go`.

tkxkd0159 (approved):

server/config/config.go
```diff
@@ -23,6 +23,9 @@ const (
 
 	// DefaultGRPCWebAddress defines the default address to bind the gRPC-web server to.
 	DefaultGRPCWebAddress = "0.0.0.0:9091"
+
+	// DefaultChanCheckTxSize defines the default maximum size of channel check tx in Baseapp
```
0Tech, line 27: Do we need the word "maximum" here? It may sound like, "you can set the size of it, but the maximum would be this value".
```suggestion
	// DefaultChanCheckTxSize defines the default size of channel check tx in Baseapp
```

server/start.go
```diff
@@ -175,6 +176,8 @@ is performed. Note, when enabled, gRPC will also be automatically enabled.
 
 	cmd.Flags().Bool(FlagPrometheus, false, "Enable prometheus metric for app")
 
+	cmd.Flags().Uint(FlagChanCheckTxSize, config.DefaultChanCheckTxSize, "The maximum size of the chan-check-tx-size")
```
0Tech, line 179: Same as above.
```suggestion
	cmd.Flags().Uint(FlagChanCheckTxSize, config.DefaultChanCheckTxSize, "The size of the chan-check-tx-size")
```

server/config/config.go
zemyblue: Yes, this is what I want. Because this is the maximum value and this is numbers greater than the set value are not assigned.

zemyblue: Oh!, I see. Since the `DefaultChanCheckTxSize` value is a default value, so you're saying it's best not to set to the highest value?

0Tech (approved):

PR #758, opened 2022-10-26
feat: add invariants to x/foundation

0Tech:
## Description

This patch will add following invariants to x/foundation:
- ModuleAccountInvariant
  - It checks whether each balances of the module accounts is in a sane state.
    - "treasury": It must be equal to the corresponding value in the pool.
- TotalWeightInvariant
  - It checks whether the total weight is in a sane state.
    - Each weight of the member is implicitly one, hence the weight must be equal to the number of the members.
- ProposalInvariant
  - It checks whether each proposal is in a sane state.
    - Old proposals must have been aborted, so they must not be in the submitted state.

## Checklist:

### DIFF
--- a/x/foundation/client/testutil/grpc.go
+++ b/x/foundation/client/testutil/grpc.go
@@ -9,7 +9,6 @@ import (
 
 	sdk "github.com/line/lbm-sdk/types"
 	"github.com/line/lbm-sdk/x/foundation"
-	stakingtypes "github.com/line/lbm-sdk/x/staking/types"
 )
 
 func (s *IntegrationTestSuite) TestGRPCParams() {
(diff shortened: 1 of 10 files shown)


zemyblue (commented):
If the invariants is broken, how can we resolve in the chain?

0Tech:
> If the invariants is broken, how can we resolve in the chain?

Halting the chain is the current short-term solution. The goal of the invariant feature is to prevent ripple effects from the violated invariant. Fixing the cause is not the goal of this feature.

x/foundation/client/testutil/grpc.go
```diff
@@ -31,7 +30,6 @@ func (s *IntegrationTestSuite) TestGRPCParams() {
 				Params: foundation.Params{
 					FoundationTax: sdk.MustNewDecFromStr("0.2"),
 					CensoredMsgTypeUrls: []string{
-						sdk.MsgTypeURL((*stakingtypes.MsgCreateValidator)(nil)),
```
zemyblue, line 34: Why does this checking remove?

0Tech: Because, the suite has removed it from the state.

zemyblue (approved):

dudong2 (approved):

PR #781, opened 2022-11-03
chore: implemented method `SetOption()`

zemyblue:
## Description

* Implemented method `SetOption()`

## Checklist:

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -118,8 +118,7 @@ func (app *BaseApp) Info(req abci.RequestInfo) abci.ResponseInfo {
 
 // SetOption implements the ABCI interface.
 func (app *BaseApp) SetOption(req abci.RequestSetOption) (res abci.ResponseSetOption) {
-	// TODO: Implement!
-	return
+	return abci.ResponseSetOption{}
 }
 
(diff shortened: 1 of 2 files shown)


dudong2 (approved):

baseapp/abci.go
```diff
@@ -118,8 +118,7 @@ func (app *BaseApp) Info(req abci.RequestInfo) abci.ResponseInfo {
 
 // SetOption implements the ABCI interface.
 func (app *BaseApp) SetOption(req abci.RequestSetOption) (res abci.ResponseSetOption) {
-	// TODO: Implement!
-	return
+	return abci.ResponseSetOption{}
```
0Tech, line 121: I have a question:
- Is there any caller of this function?
  - If there is, what's the improvement by this change?

zemyblue: `SetOption` is ABCI interface to support the function to set the real-time option in the application layer through the cli. You can find the example in https://docs.tendermint.com/v0.34/app-dev/abci-cli.html#counter-another-example.

But, since lbm-sdk(same with cosmos-sdk) does not have  a function that supports setting options in real-time, there is no problem in implementing the default response value without needing to use it.

0Tech (approved):

PR #784, opened 2022-11-07
fix: make field key matching in x/token & x/collection case-sensitive

0Tech:
## Description

This patch would make the field key matching in x/token & x/collection case-sensitive.

## Checklist:


x/collection/msgs_test.go
```diff
@@ -2,9 +2,11 @@ package collection_test
 
 import (
 	"fmt"
-	"github.com/line/lbm-sdk/x/auth/legacy/legacytx"
+	"strings"
 	"testing"
 
+	"github.com/line/lbm-sdk/x/auth/legacy/legacytx"
```
zemyblue, line 8: Please, gather this import into the `line/lbm-sdk` group below.

zemyblue (approved):

x/collection/keeper/supply.go
```diff
@@ -317,19 +317,20 @@ func (k Keeper) ModifyContract(ctx sdk.Context, contractID string, operator sdk.
...
 
-	modifiers := map[string]func(string){
-		collection.AttributeKeyName.String(): func(name string) {
+	modifiers := map[collection.AttributeKey]func(string){
+		collection.AttributeKeyName: func(name string) {
 			contract.Name = name
 		},
-		collection.AttributeKeyBaseImgURI.String(): func(uri string) {
+		collection.AttributeKeyBaseImgURI: func(uri string) {
 			contract.Name = uri
```
loin3, line 325:
```suggestion
			contract.BaseImgUri = uri
```
Shouldn't this be changed like this?

0Tech: Thanks. I didn't notice that. I should revisit entire modify logics on x/token and x/collection, and add the corresponding tests. It would be another bug fix PR.

x/collection/msgs.go
```diff
@@ -166,26 +166,26 @@ func ValidatePermission(permission Permission) error {
...
+		AttributeKeyMeta.String(): validateMeta,
 	}
 
 	return validateChange(change, validators)
 }
 
-func validateChange(change Attribute, validators map[AttributeKey]func(string) error) error {
-	validator, ok := validators[AttributeKeyFromString(change.Key)]
+func validateChange(change Attribute, validators map[string]func(string) error) error {
+	validator, ok := validators[change.Key]
```
tkxkd0159, line 188: Is there any reason to change like this? Overall, isn't it more consistent to use the key as the attribute key data type? I don't know the context of this changes so hope to understand more clearly.

0Tech: Using the keys makes the logic case insensitive. The return type of the conversion does not include any errors, but I didn't want to change the return types, because it's a deprecated module.

tkxkd0159 (approved):

0Tech:
CI fails due to the latest changes of workflow in rc9/v0.46.0.

PR #793, opened 2022-11-10
feat: enable to use libsodium version ostracon

ulbqb:
## Description

This PR enable to use libsodium version ostracon. We can use libsodium version ostracon by the following command.
```bash
$ LBM_BUILD_OPTIONS=libsodium make install      
```

## Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -138,7 +138,7 @@ jobs:
           USE_PRELOAD: 1,4
           SAVE_BRANCH_LAUNCH_DEPTH: 1
         run: |
-          cat pkgs.txt.part.${{ matrix.part }} | xargs go test -mod=readonly -timeout 30m -coverprofile=${{ matrix.part }}profile.out -covermode=atomic -tags='norace ledger test_ledger_mock goleveldb'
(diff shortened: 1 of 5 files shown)


tnasu:
Would you add the test with libsodium on `github actions`?

ulbqb:
> Would you add the test with libsodium on github actions?

@tnasu OK, I'll add test. However, adding tests does not guarantee that ostracon is using libsodium. There seem to be no way to tell if ostracon is using libsodium.

tnasu:
@ulbqb Yes, it does. `ostracon` doesn't tell us whether to use `libsodium` and what kind of tm-db-backend it uses. It's not convenient. (Thanks for your creating a issue: https://github.com/line/ostracon/issues/486)

But, `lbm-sdk` is tested with `goleveldb`, so I recommend adding the test with `goleveldb` and `libsodium`.

zemyblue (commented):
I think this change need to be applied in lbm-sdk v0.46.0-rcx. So could you change the target branch to `rc9/v0.46.0`?

ulbqb:
@zemyblue 
OK. But changing the target branch broke this PR, so I force-pushed.

tnasu (approved):

PR #821, opened 2022-12-06
feat: Get validator pubkey considering KMS

kokeshiM0chi:
## Description

Change to get validator pubkey considering KMS.

However, the LBM coupled with the Ostracon only allows the local private validator without the remote private validator support:


server/oc_cmds.go
```diff
@@ -5,15 +5,17 @@ package server
...
 	pvm "github.com/line/ostracon/privval"
+	"github.com/line/ostracon/types"
 	ostversion "github.com/line/ostracon/version"
 	"github.com/spf13/cobra"
 	yaml "gopkg.in/yaml.v2"
-
-	"github.com/line/lbm-sdk/client"
-	cryptocodec "github.com/line/lbm-sdk/crypto/codec"
-	sdk "github.com/line/lbm-sdk/types"
 )
```
zemyblue, line 19: Please follow the rules for separating imports by group. 
1. go library (ex: os, rand, fmt)
2. other github path (ex: github.com/spf13/cobra)
3. line github path (ex: github.com/line/ostracon/node)
4. current path (ex: github.com/line/lbm-sdk/types)

zemyblue (commented):
* Please add changelog in `CHANGELOG.md`. 
* Please add a unittest to check your modifications.

server/oc_cmds.go
kokeshiM0chi: I fixed as indicated.

kokeshiM0chi:
@zemyblue 
I've fixed these below. PTAL

> * ase add changelog in `CHANGELOG.md`.
https://github.com/line/lbm-sdk/pull/821/commits/b8a1a9afed5f7c4e8a29efef1675a1d21c86cdf7

> * Please add a unittest to check your modifications.
https://github.com/line/lbm-sdk/pull/821/commits/858b05a5db7a402b34b6a0e46e26470bb2a4cd7d

server/oc_cmds.go
```diff
@@ -43,29 +47,64 @@ func ShowValidatorCmd() *cobra.Command {
...
+		if err != nil {
+			return err
+		}
+		serverCtx := GetServerContextFromCmd(cmd)
+		log := serverCtx.Logger
+		pv, err = node.CreateAndStartPrivValidatorSocketClient(config.PrivValidatorListenAddr, chainID, log)
+		if err != nil {
+			return err
+		}
+	} else {
```
zemyblue, line 70: Why does this `showValidator` function create private key?
I think creating private key seems to go against this function rule.

server/oc_cmds.go
```diff
@@ -43,29 +47,64 @@ func ShowValidatorCmd() *cobra.Command {
...
+	bz, err := osjson.Marshal(pubKey)
+	if err != nil {
+		return fmt.Errorf("failed to marshal private validator pubkey: %w", err)
+	}
+
+	fmt.Println(string(bz))
+	return nil
+}
+
+func LoadChainID(config *cfg.Config) (string, error) {
```
torao, line 92: This function seems to be used only in this file. Why does it need to be public?
(You might call a private function in the same package, even for testing purposes.)

kokeshiM0chi: I'm at a loss as well. Packages are separate for test files and others, and this test belongs to the `server_test` package. And this is tested by calling a private function in `util_test.go`. Therefore, this function could not be called from the test function unless it was made public.

torao: If the `server_test` package is used for that reason, I think it would be better to make the private functions in `util_test.go` public,  if the impact isn't too large.

kokeshiM0chi: It's certainly a good idea to make the test function public. Thanks for your suggestion.

server/oc_cmds.go
kokeshiM0chi: This is the same process as [Ostracon](https://github.com/line/ostracon/blob/v1.0.7/cmd/ostracon/commands/show_validator.go#L34), and when I checked with @torao, it seems that a key is created to encrypt communication between Ostracon and KMS.

zemyblue (approved):

torao (approved):
LGTM

Issue #683, opened 2022-09-21
Add API to query account sequence number in mempool


zemyblue:
## Summary

In order to know the appropriate the sequence number of the account, we need to be able to inquire the sequence number of the account registered in the mempool. 

## Problem Definition

If some target node is down, we need to wait or change the target node. 
However, if the transaction created by the account exists in the mempool in the changed target node when changing the target node, the appropriate sequence number of the account cannot be retrieved.

### Background
* The normal query api for the account info is inquired from `deliverState`.
* This `deliverState` is updated when the block is confirmed. So we cannot query the information to be reflected in mempool through the normal query api. 

So we need to be able to query the sequence number of the account in the mempool.

## Proposal

Add the query api about account info from `checkState`.

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #840, opened 2022-12-23
feat: enable querying based on mempool status (only gRPC)

tkxkd0159:
## Description

closes: #683 

To avoid conflict with upstream, make another `createQueryContext` function.

## Checklist:


tkxkd0159:
`codecov/patch` said new added lines were not tested but it's not true because they were tested by `IntegrationTestSuite`. Those lines were hit when I executed tests at the local machine.

baseapp/abci.go
```diff
@@ -920,3 +921,19 @@ func splitPath(requestPath string) (path []string) {
...
+func (app *BaseApp) createQueryContextWithCheckState() sdk.Context {
+
+	cacheMS := app.checkState.CacheMultiStore()
+
+	// branch the commit-multistore for safety
+	app.checkStateMtx.RLock()
+	ctx := sdk.NewContext(
+		cacheMS, app.checkState.ctx.BlockHeader(), true, app.logger,
+	).WithMinGasPrices(app.minGasPrices).WithBlockHeight(app.LastBlockHeight())
+	app.checkStateMtx.RUnlock()
```
jaeseung-bae, line 936: Can you move "app.checkStateMtx.RUnlock()" to right after "app.checkStateMtx.RLock()" with "defer" keyword.
If you move "defer app.checkStateMtx.RUnlock()" to L933, Mutex gets closed even if any exceptional case happens.

tkxkd0159: Thank you for your comment but the logic between `RLock` is deterministic. So it's impossible to occur any exceptional cases which makes deadlock.

jaeseung-bae: Ok, I understand.

jaeseung-bae:
LGTM

jaeseung-bae (approved):
LGTM

loin3 (approved):

PR #849, opened 2023-01-03
fix!: introduce codespace into x/collection,token

0Tech:
## Description

This patch would import codespace for x/collection,token errors, and make the modules to emit the exactly same errors as daphne does.

Note: It would affect the performance of the target modules.

## Motivation and context

Our client requires it.

## Checklist:


types/errors/link.go
```diff
@@ -0,0 +1,12 @@
+package errors
```
zemyblue, line 1: How about changing the file name to `legacy_error.go` or others, not `link.go`?

types/errors/link.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	linkCodespace = "link"
+)
+
+var (
+	// ErrError             = Register(linkCodespace, 1, "error")
```
zemyblue, line 8: This comment seems to be a bug or test code. 
How about removing it or adding the reason comment above it?

x/collection/errors.go
```diff
@@ -0,0 +1,58 @@
+package collection
+
+import (
+	sdkerrors "github.com/line/lbm-sdk/types/errors"
+)
+
+const collectionCodespace = ModuleName
+
+var (
+	// ErrTokenExist                    = sdkerrors.Register(collectionCodespace, 1, "token symbol, token-id already exists")
```
zemyblue, line 10: dito

x/collection/keeper/grpc_query.go
```diff
@@ -85,6 +85,16 @@ func (s queryServer) AllBalances(c context.Context, req *collection.QueryAllBala
...
 		balances = append(balances, coin)
+
+		if err := collection.ValidateNFTID(tokenID); err == nil {
+			s.keeper.iterateDescendants(ctx, req.ContractId, tokenID, func(tokenID string, _ int) (stop bool) {
+				coin := collection.NewCoin(tokenID, sdk.OneInt())
+				balances = append(balances, coin)
+
+				return false
+			})
+		}
```
zemyblue, line 96: Why is this iteration necessary?

types/errors/link.go
0Tech: These errors would be a spec, and last as long as x/token and x/collection would. So I think it should be fine as is (I don't like these errors though).

types/errors/link.go
0Tech: OK, I will remove them.

types/errors/link.go
0Tech: I added a comment above these errors.

x/collection/keeper/grpc_query.go
0Tech: Because daphne returns the whole nfts including the descendants. The DB of finschia only cares about the root nft for the performance and other reasons.

0Tech: It could be better to apply it on another PR.

types/errors/link.go
```diff
@@ -0,0 +1,11 @@
+package errors
+
+const (
+	linkCodespace = "link"
```
zemyblue, line 4: Why is the codespace `link`?

0Tech: Because it was "link" in daphne. For the backward compatibility, we should use the same codespace.

x/collection/keeper/grpc_query.go
0Tech: Removed the fix.

x/collection/keeper/nft.go
```diff
@@ -4,15 +4,31 @@ import (
...
+
+	if err2 := collection.ValidateFTID(tokenID); err2 != nil {
+		return collection.ErrTokenNotExist.Wrap(tokenID)
+	}
+	classID := collection.SplitTokenID(tokenID)
+	if _, err2 := k.GetTokenClass(ctx, contractID, classID); err2 != nil {
+		return collection.ErrTokenNotExist.Wrap(tokenID)
+	}
+
+	return collection.ErrTokenNotNFT.Wrap(tokenID)
```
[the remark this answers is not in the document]
0Tech: I don't disagree with your opinion, so I may commit your suggestion. Just out of curiosity, what's the expected optimization?

simapp/app.go
```diff
@@ -313,9 +314,9 @@ func NewSimApp(
 	foundationConfig := foundation.DefaultConfig()
 	app.FoundationKeeper = foundationkeeper.NewKeeper(appCodec, keys[foundation.StoreKey], app.BaseApp.MsgServiceRouter(), app.AccountKeeper, app.BankKeeper, authtypes.FeeCollectorName, foundationConfig, foundation.DefaultAuthority().String())
 
-	classKeeper := classkeeper.NewKeeper(appCodec, keys[class.StoreKey])
-	app.TokenKeeper = tokenkeeper.NewKeeper(appCodec, keys[token.StoreKey], app.AccountKeeper, classKeeper)
-	app.CollectionKeeper = collectionkeeper.NewKeeper(appCodec, keys[collection.StoreKey], app.AccountKeeper, classKeeper)
+	app.ClassKeeper = classkeeper.NewKeeper(appCodec, keys[class.StoreKey])
+	app.TokenKeeper = tokenkeeper.NewKeeper(appCodec, keys[token.StoreKey], app.AccountKeeper, app.ClassKeeper)
+	app.CollectionKeeper = collectionkeeper.NewKeeper(appCodec, keys[collection.StoreKey], app.AccountKeeper, app.ClassKeeper)
```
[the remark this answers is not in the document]
0Tech: I exposed the keeper just for the tests on the error codes. FYI, we made a decision not to create its own module because the expected module's name was `contract`.

x/token/keeper/send.go
```diff
@@ -21,10 +21,11 @@ func (k Keeper) Send(ctx sdk.Context, contractID string, from, to sdk.AccAddress
 
 func (k Keeper) AuthorizeOperator(ctx sdk.Context, contractID string, holder, operator sdk.AccAddress) error {
 	if _, err := k.GetClass(ctx, contractID); err != nil {
-		return err
+		panic(err)
```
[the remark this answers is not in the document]
0Tech: Because the caller must ensure the condition is true. I may remove the entire check logic if it looks strange.

x/collection/keeper/grpc_query_test.go
```diff
@@ -81,7 +81,7 @@ func (s *KeeperTestSuite) TestQueryAllBalances() {
 			address:    s.customer,
 			valid:      true,
 			postTest: func(res *collection.QueryAllBalancesResponse) {
-				s.Require().Equal(s.numNFTs+1, len(res.Balances))
+				s.Require().Equal(2+1+1, len(res.Balances))
```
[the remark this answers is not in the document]
0Tech: Thank you, I made the code more readable.

x/collection/keeper/nft.go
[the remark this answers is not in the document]
tkxkd0159: it's very trivial (register jump). Go's compiler may optimize this automatically but I just suggested this because the conditions are orthogonal. So you can ignore it.

x/token/keeper/send.go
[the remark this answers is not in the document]
tkxkd0159: I thought that caller may be able to input contractID incorrectly sometimes because of human error. And the method such as `TransferFT` returns error instead of panic. So I guess every msg which consume `contractID` should have same flow for consistency.

x/collection/keeper/nft.go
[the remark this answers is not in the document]
jaeseung-bae: As Jaden said, Go's compiler may optimize.
In my opinion, the previous one(single if error handling) looks betters because the code follows go's error-handling convention. 
I don't need to memorize the previous error check condition with previous one(`if error check`), but with new code(`if, else if, else if`) I would try to memorize the previous one which is unnecessary though. 
When I add other error check logic, I'll definitely look up all `if, else if, else if` which is not good.

x/collection/keeper/nft.go
```diff
@@ -4,15 +4,31 @@ import (
 	gogotypes "github.com/gogo/protobuf/types"
 
 	sdk "github.com/line/lbm-sdk/types"
-	sdkerrors "github.com/line/lbm-sdk/types/errors"
 	"github.com/line/lbm-sdk/x/collection"
 )
 
+func legacyNFTNotFoundError(k Keeper, ctx sdk.Context, contractID string, tokenID string) error {
+	// legacy
+	if err2 := collection.ValidateLegacyNFTID(tokenID); err2 == nil {
```
jaeseung-bae, line 12: `err2 == nil` is intentional, isn't it?
When I look at this first time, I thought it was a bug because we use `err != nil` a lot by convention.

Sorry for that I don't have any idea. Is there any way to express it in another way?

x/collection/keeper/nft.go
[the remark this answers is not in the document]
0Tech: > In my opinion, the previous one(single if error handling) looks better because the code follows go's error-handling convention.

It's the reason behind this logic. Thanks for the clarification.

> As Jaden said, Go's compiler may optimize.

I don't see any practical optimization point. It would be almost same with or without the `if else` statements. Let the machines do what they do well, and we should concentrate on the human's job, like the readability. If we can come up with some trivial optimization, they can too.

x/collection/keeper/nft.go
0Tech: We can change it into `err2 != nil` and use an `else` block, but it's against the common coding style.
As a last resort, I will leave a comment.

x/collection/keeper/nft.go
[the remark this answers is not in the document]
jaeseung-bae: I agree with going for readability :)

x/token/keeper/send.go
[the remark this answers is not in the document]
0Tech: Generally, you are correct. However, we must always check the existence when we get `contractID` as a parameter. So, the focus should be on "what is the interface". Unfortunately, the function is, as of now, public.
The correct way to fix this is, introducing `internal` and hide the private functions. However, this module is deprecated one, so we won't spend any time refactoring it.
FYI: The function has been made public for black-box testing.

jaeseung-bae (approved):

tkxkd0159 (approved):

PR #855, opened 2023-01-05
test: add tests for FeePool and handler

jaeseung-bae:
## Description

* Add unit tests for uncovered statements in distribution module
  * Testcases for handler and FeePool

## Checklist:


Scottbae37:
@tkxkd0159 
Thanks for the comments.
 I've found no handler in the latest cosmos-sdk and also they seemed test for DistributeFromFeePool.
The tests I added cover only one additional case which is an exceptional case of 'types.ErrBadDistribution'.

I agree, following cosmos-sdk is more expensive than adding a test for only one additional case.
I'll close this PR.

jaeseung-bae:
> In my opinion, this PR should be closed. There are some reasons
> 
> 1. we already don't use handler anymore. we registered all gRPC router and every tx are routed by them.
> 2. test cases related to `SetFeePool` already set in upstream. so we should follow this to avoid future conflicts & redundant test

Thanks for the comments.
I've found no handler in the latest cosmos-sdk and also they seemed test for DistributeFromFeePool.
The tests I added cover only one additional case which is an exceptional case of 'types.ErrBadDistribution'.

I agree, following cosmos-sdk is more expensive than adding a test for only one additional case.
I'll close this PR.

Issue #861, opened 2023-01-11
Remove max_gas verification from lbm-sdk


0Tech:
- Get ready for line/ostracon#507
  - Remove redundant logic in lbm-sdk

PR #673, opened 2022-09-14
fix: fix max gas validation

0Tech:
## Description

This patch fixes bugs relevant to #48. We should revisit this and #48. Please refer to tendermint/tendermint#6740.

Related: line/ostracon#158

## Checklist:

- [x] I followed the [contributing guidelines](https://github.com/line/lbm-sdk/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/line/lbm-sdk/blob/main/CODE_OF_CONDUCT.md).
- [x] I have added a relevant changelog to `CHANGELOG.md`
- [ ] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated API documentation `client/docs/swagger-ui/swagger.yaml`

PR #895, opened 2023-02-13
fix: remove max gas validation

0Tech:
## Description

closes: #861

## Checklist:

### DIFF
--- a/baseapp/abci.go
+++ b/baseapp/abci.go
@@ -175,8 +175,7 @@ func (app *BaseApp) BeginBlock(req ocabci.RequestBeginBlock) (res abci.ResponseB
 	if app.checkState != nil {
 		app.checkState.ctx = app.checkState.ctx.
 			WithBlockGasMeter(gasMeter).
-			WithHeaderHash(req.Hash).
-			WithConsensusParams(app.GetConsensusParams(app.deliverState.ctx))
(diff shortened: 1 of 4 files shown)


jaeseung-bae (dismissed):

baseapp/abci.go
```diff
@@ -175,8 +175,7 @@ func (app *BaseApp) BeginBlock(req ocabci.RequestBeginBlock) (res abci.ResponseB
 	if app.checkState != nil {
```
jaeseung-bae, line 175: The comments below seem they need an update. I'm not sure. Can you check?

`// we also set block gas meter to checkState in case the application needs to`
`// verify gas consumption during (Re)CheckTx`

0Tech: The comment states about gas meter, not max_gas, so it would be OK to leave it as is.

jaeseung-bae: Ok~

jaeseung-bae (approved):

zemyblue (commented):
Does this PR roll back #673 ?

0Tech:
> Does this PR roll back #673 ?

Yes, but partially. #673 does set consensus params to the deliver state. The upstream also sets these. So the relevant logic and tests would remain.

zemyblue (approved):

Issue #899, opened 2023-02-14
MsgModify in x/token,collection should also accept "uri"


0Tech:
## Summary

MsgModify accepts `base_img_uri` or `img_uri` for backward-compatibility. MsgModify should also accept `uri`, because the corresponding proto has been updated into `uri`.

## Problem Definition

The current implementation may confuse new users, so we should add a key `uri`.

## Proposal

1. Make `MsgModify` to accept a key `uri` for its changes.
2. Users must use either of `uri` or old key. If they use both of the key in the same time, the validation logic must emit an error.

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

zemyblue:
If both `img_uri` and `uri`are exited in the same Msg, is it confused?

PR #900, opened 2023-02-14
feat: add "uri" for MsgModify and deprecate the old ones

0Tech:
## Description

closes: #899

## Checklist:


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * (client) [\#890](https://github.com/line/lbm-sdk/pull/890) Map Ostracon:ErrTxInMap to lbm-sdk:ErrTxInMempoolCache
 * (x/collection) [\#894](https://github.com/line/lbm-sdk/pull/894) Change the default params of x/collection
 * (ante) [\#895](https://github.com/line/lbm-sdk/pull/895) Remove max gas validation
+* (ante) [\#900](https://github.com/line/lbm-sdk/pull/900) Add uri for MsgModify and deprecate the old ones
```
0Tech, line 54:
```suggestion
* (x/collection,token) [\#900](https://github.com/line/lbm-sdk/pull/900) Add uri for MsgModify and deprecate the old ones
```

0Tech:
I will revise the relevant event.
as-is: emit uri or img_uri
to-be: emit both of uri and img_uri

go.mod
```diff
@@ -102,7 +102,7 @@ require (
 	github.com/minio/highwayhash v1.0.2 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/mtibben/percent v0.2.1 // indirect
-	github.com/onsi/gomega v1.20.0 // indirect
+	github.com/onsi/gomega v1.22.1 // indirect
```
zemyblue, line 105: Why is this changed? Do you know?

zemyblue:
> 

Please check if no problem with related team, @0Tech .

go.mod
0Tech: Because our proto script runs `go mod tidy`.

x/token/validation.go
```diff
@@ -97,3 +98,13 @@ func validateChange(change Attribute) error {
 func ValidateContractID(id string) error {
 	return class.ValidateID(id)
 }
+
+func canonicalKey(key string) string {
+	convert := map[string]string{
+		AttributeKeyImageURI.String(): AttributeKeyURI.String(),
+	}
```
dudong2, line 105: AttributeKeyImageURI and AttributeKeyURI are all fixed values, so convert also does not change. It is unnecessary to declare this every time. How about pre-declaring it with the var at the top?

0Tech: We should avoid using globals if it's not performance-critical. Also, there is no other place using this map.
Could you elaborate your concern on it?

0Tech:
> > 
> 
> Please check if no problem with related team, @0Tech .

Thank you.

For other reviewers: this PR is now under review by our clients.

0Tech:
> > > 
> > 
> > 
> > Please check if no problem with related team, @0Tech .
> 
> Thank you.
> 
> For other reviewers: this PR is now under review by our clients.

Done.

proto/lbm/collection/v1/tx.proto
```diff
@@ -466,7 +466,7 @@ message MsgModify {
   // Note: if token type is of FTs, the index cannot be empty.
   string token_index = 4;
   // changes to apply.
-  // possible attribute keys on modifying collection: name, base_img_uri, meta.
+  // possible attribute keys on modifying collection: name, uri, base_img_uri (deprecated), meta.
```
zemyblue, line 469: I think we don't need `base_image_uri` for now. Isn't it?

proto/lbm/token/v1/tx.proto
```diff
@@ -274,7 +274,7 @@ message MsgModify {
   // the address of the grantee which must have modify permission.
   string owner = 2;
   // changes to apply.
-  // possible attribute keys are: name, img_uri, meta
+  // possible attribute keys are: name, uri, img_uri (deprecated), meta
```
zemyblue, line 277: Can the `img_url` key be used after this PR merge?

proto/lbm/collection/v1/tx.proto
0Tech: The client didn't mention about tx. So I have no idea whether we can remove it.

proto/lbm/token/v1/tx.proto
0Tech: In the current implementation, yes.

zemyblue (approved):

jaeseung-bae (approved):
LGTM

x/collection/msgs.go
```diff
@@ -221,6 +222,16 @@ func validateChange(change Attribute, validators map[string]func(string) error)
 	return validator(change.Value)
 }
 
+func canonicalKey(key string) string {
+	convert := map[string]string{
```
loin3, line 226: Is the use of a map rather than a simple if statement to make it easier to respond when another change occurs later? For example if `Meta` is deprecated and `Metadata` should be used then just add `AttributeKeyMeta.String(): AttributeMetadata.String(),` ?

0Tech: Exactly.

loin3 (approved):

Issue #910, opened 2023-02-23
Move controls of CreateValidatorAuthorization from x/foundation to x/gov


0Tech:
After changing Finschia from a private chain to a consortium chain, 
controls of CreateValidatorAuthorization need to be transferred from Foundation members(x/foundation) to Validators(x/gov).

But the timing of this transfer is uncertain for now, and the final decision should be made by Foundation later.

So the trigger logic is required, which means the transfer should be activated after a trigger executed by foundation members (x/foundation).

+ remove CreateValidatorAuthorization is also required under both controls, x/foundation and x/gov. (for being a public chain later)

Issued by the planning team.

PR #912, opened 2023-02-23
feat: introduce censorship into x/foundation

0Tech:
## Description

This patch would:
- introduce `Censorship`, which has its target message type url and authority.
  - its authority must be one of the following enum values:
    - `CENSORSHIP_AUTHORITY_GOVERNANCE`
    - `CENSORSHIP_AUTHORITY_FOUNDATION`
  - the authority can trigger the following messages on its target message type url:
    - `/lbm.foundation.v1.MsgUpdateCensorship`
    - `/lbm.foundation.v1.MsgGrant`
    - `/lbm.foundation.v1.MsgRevoke`
- introduce `/lbm.foundation.v1.MsgUpdateCensorship`, which updates censorship information.
  - the corresponding enum value of the new authority must be lesser than the current one.
  - one may remove (or stop) the censorship by setting its authority to `CENSORSHIP_AUTHORITY_UNSPECIFIED`.
- introduce `FoundationExecProposal` which allows x/gov's authority to trigger x/foundation's messages.
  - the signer(s) of the message MUST be x/gov's authority.
  - For now, only the following messages are allowed:
    - `/lbm.foundation.v1.MsgUpdateCensorship`
    - `/lbm.foundation.v1.MsgGrant`
    - `/lbm.foundation.v1.MsgRevoke`

closes: #910

## Checklist:


x/foundation/README.md
```diff
@@ -441,9 +452,32 @@ The messages that are part of this proposal won't be executed if:
 * the proposal has not been accepted by the decision policy.
 * the proposal has already been successfully executed.
 
+## Msg/UpdateCensorship
+
+A censorship information is updated by using the `MsgUpdateCensorship` message.
+One cannot introduce a censorship over a new message type URL by this message.
+
++++ https://github.com/line/lbm-sdk/blob/main/proto/lbm/foundation/v1/tx.proto#L215-L222
```
0Tech, line 460: Update into permlink after this PR merged.

x/foundation/client/cli/query.go
```diff
@@ -356,6 +357,35 @@ func NewQueryCmdTallyResult() *cobra.Command {
...
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientQueryContext(cmd)
+			if err != nil {
+				return err
+			}
+			queryClient := foundation.NewQueryClient(clientCtx)
+
+			censorships := foundation.QueryCensorshipsRequest{}
+			res, err := queryClient.Censorships(context.Background(), &censorships)
```
0Tech, line 375: No pagination?

x/foundation/msgs.go
```diff
@@ -445,6 +445,42 @@ func (m MsgLeaveFoundation) GetSignBytes() []byte {
...
+func (m MsgUpdateCensorship) Type() string {
+	return sdk.MsgTypeURL(&m)
+}
+
+// Route implements the LegacyMsg.Route method.
+func (m MsgUpdateCensorship) Route() string {
+	return sdk.MsgTypeURL(&m)
+}
+
+// GetSignBytes implements the LegacyMsg.GetSignBytes method.
```
0Tech, line 479: Add a test on amino.

x/foundation/README.md
```diff
@@ -205,11 +207,19 @@ Proposals are pruned:
 
 whichever happens first.
 
-## Authorization
+## Censorship
 
 The foundation module defines interfaces of authorizations on messages to
 enforce_censorship_ on its execution. The other modules may deny the execution
```
0Tech, line 213:
```suggestion
enforce _censorship_ on its execution. The other modules may deny the execution
```

typo.

x/foundation/README.md
0Tech: It would be addressed by #913.

dudong2 (approved):
LGTM

dudong2 (dismissed):

x/foundation/keeper/keys.go
```diff
@@ -19,7 +19,8 @@ var (
 	proposalByVPEndKeyPrefix = []byte{0x13}
 	voteKeyPrefix            = []byte{0x14}
 
-	grantKeyPrefix = []byte{0x20}
+	censorshipKeyPrefix = []byte{0x20}
```
jaeseung-bae, line 22: I'm just wondering conflict between `censorshipKeyPrefix=0x20` and `grantKeyPrefix=0x20`(previously used).
Is there any possibility that old grant data with grantKeyPrefix(0x20) gets loaded and failed to interpret that as censorship data?

0Tech: Technically, you are right about it. However, AFAIK, this sdk is only used by our nodes, and we won't simply change the binary. It would not allowed to apply this kind of change after the release, but it seems OK now.

jaeseung-bae: Ok, I understand

jaeseung-bae (approved):

tkxkd0159 (approved):
LGTM

zemyblue (commented):
How can we move the auth of `CreateValidatorAuthorization` from `x/foundation` to `x/gov`?

0Tech:
> How can we move the auth of `CreateValidatorAuthorization` from `x/foundation` to `x/gov`?

Please refer to the corresponding item in the description of this PR:

> introduce `/lbm.foundation.v1.MsgUpdateCensorship`, which updates censorship information.

You can change the authority into `CENSORSHIP_AUTHORITY_GOVERNANCE`.

x/foundation/keeper/censorship.go
```diff
@@ -6,6 +6,72 @@ import (
...
+	url := censorship.MsgTypeUrl
+
+	oldCensorship, err := k.GetCensorship(ctx, url)
+	if err != nil {
+		return err
+	}
+
+	newAuthority := censorship.Authority
+	oldAuthority := oldCensorship.Authority
+	if newAuthority >= oldAuthority {
```
zemyblue, line 33: Would this need the same case ?

zemyblue (commented):
Do you mean we need to use `MsgUpdateCencorship` if you want to move the auth. 
But the autht to want to change should be lower the enum number. Is it right?

x/foundation/keeper/censorship.go
0Tech: So, you cannot raise the authority, like x/gov -> x/foundation.

0Tech: > Do you mean we need to use `MsgUpdateCencorship` if you want to move the auth. But the autht to want to change should be lower the enum number. Is it right?

So, you cannot raise the authority, like x/gov -> x/foundation.

0Tech: > Would this need the same case ?

The equality means, the msg has no changes, so it rejects.

loin3 (approved):

x/foundation/keeper/member.go
```diff
@@ -160,6 +162,23 @@ func (k Keeper) validateAuthority(authority string) error {
...
+		return err
+	}
+
+	authorityAddrs := map[foundation.CensorshipAuthority]string{
+		foundation.CensorshipAuthorityGovernance: authtypes.NewModuleAddress(govtypes.ModuleName).String(),
+		foundation.CensorshipAuthorityFoundation: k.authority,
+	}
+	if expected := authorityAddrs[censorship.Authority]; authority != expected {
+		return sdkerrors.ErrUnauthorized.Wrapf("invalid authority; expected %s, got %s", expected, authority)
+	}
```
zemyblue, line 177: I think it's good to use `if-else`, how about this?

PR #946, opened 2023-03-29
fix: fix broken x/foundation invariant on treasury

0Tech:
## Description

This patch would truncate the tax before updating its own pool, ensuring the corresponding balance and the pool are in sync.

## Checklist:

### DIFF
--- a/x/foundation/keeper/internal/treasury.go
+++ b/x/foundation/keeper/internal/treasury.go
@@ -6,8 +6,9 @@ import (
 	"github.com/line/lbm-sdk/x/foundation"
 )
 
-func (k Keeper) CollectFoundationTax(ctx sdk.Context, from sdk.AccAddress) error {
-	feesCollectedInt := k.bankKeeper.GetAllBalances(ctx, from)
+func (k Keeper) CollectFoundationTax(ctx sdk.Context) error {
+	feeCollector := k.authKeeper.GetModuleAccount(ctx, k.feeCollectorName).GetAddress()
+	feesCollectedInt := k.bankKeeper.GetAllBalances(ctx, feeCollector)
 	feesCollected := sdk.NewDecCoinsFromCoins(feesCollectedInt...)
 
 	// calculate the tax
@@ -15,7 +16,7 @@ func (k Keeper) CollectFoundationTax(ctx sdk.Context, from sdk.AccAddress) error
 	tax, _ := feesCollected.MulDecTruncate(taxRatio).TruncateDecimal()
 
 	// collect the tax
-	if err := k.FundTreasury(ctx, from, tax); err != nil {
+	if err := k.FundTreasury(ctx, feeCollector, tax); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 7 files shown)


x/foundation/keeper/internal/treasury.go
```diff
@@ -6,27 +6,16 @@ import (
...
-	pool.Treasury = pool.Treasury.Add(tax...)
-	k.SetPool(ctx, pool)
-
-	// collect tax to the foundation treasury
-	amount, _ := tax.TruncateDecimal()
-	if err := k.bankKeeper.SendCoinsFromModuleToModule(ctx, k.feeCollectorName, foundation.TreasuryName, amount); err != nil {
-		return err
-	}
+	// collect the tax
+	k.FundTreasury(ctx, from, tax)
```
loin3, line 18: There is no error handling here. Is it ok?
```suggestion
	err := k.FundTreasury(ctx, from, tax)
	if err != nil {
		return err
	}
```

x/foundation/keeper/internal/treasury.go
```diff
@@ -6,27 +6,16 @@ import (
 	"github.com/line/lbm-sdk/x/foundation"
 )
 
-func (k Keeper) CollectFoundationTax(ctx sdk.Context) error {
-	// fetch and clear the collected fees for the fund, since this is
-	// called in BeginBlock, collected fees will be from the previous block
-	feeCollector := k.authKeeper.GetModuleAccount(ctx, k.feeCollectorName)
-	feesCollectedInt := k.bankKeeper.GetAllBalances(ctx, feeCollector.GetAddress())
+func (k Keeper) CollectFoundationTax(ctx sdk.Context, from sdk.AccAddress) error {
+	feesCollectedInt := k.bankKeeper.GetAllBalances(ctx, from)
```
zemyblue, line 10: Why do you extract the feeCollectorModule address out of this function?
I think this function doesn't know the `from` is the address of collector module. 
And it seems to open up the ability to insert other addresses as `from` except collection module.

x/foundation/keeper/internal/treasury.go
zemyblue, line 18: The `from` is module address but the internal `FundTreasury` function is treated as an account address. Is it no problem?

zemyblue: It's already converted to an address, so that shouldn't be a problem. I see. Is this your intention?

0Tech: Yes, my intention is using `FundTreasury()` here so I can remove the duplicated logic.

x/foundation/keeper/internal/treasury.go
0Tech: It does not allow the app developers to use another address, because the function `CollectFoundationTax()` is not API of this module. So, one must forge this package to do the bad things.
But I don't mind to revert this change, so I will revert it.

zemyblue (approved):

loin3 (approved):

PR #953, opened 2023-04-07
fix: allow zero amount of coin in x/collection Query/Balance

0Tech:
## Description

This patch would allow zero amount of coin in x/collection Query/Balance.

## Checklist:


loin3:
in test code, newCoin is still used. is it ok?
https://github.com/line/lbm-sdk/blob/main/x/collection/keeper/grpc_query_test.go#L30

0Tech:
> in test code, NewCoin is still used. is it ok? https://github.com/line/lbm-sdk/blob/main/x/collection/keeper/grpc_query_test.go#L30

Technically, it would be OK, because the balance would not be zero, but it would be better not to use NewCoins() in the test. I will apply your suggestion. Thank you.

zemyblue (approved):

loin3 (approved):

PR #961, opened 2023-04-11
fix: do not loop enum in x/collection

0Tech:
## Description

This patch would fix possible non-determinism.
It also removes comments which has vulnerable logic.

## Checklist:

### DIFF
--- a/x/collection/keeper/supply.go
+++ b/x/collection/keeper/supply.go
@@ -26,6 +26,7 @@ func (k Keeper) CreateContract(ctx sdk.Context, creator sdk.AccAddress, contract
 		Grantee:    creator.String(),
 	}
 	ctx.EventManager().EmitEvent(collection.NewEventGrantPermTokenHead(eventGrant))
+	// 0 is "unspecified"
 	for i := 1; i < len(collection.Permission_value); i++ {
 		p := collection.Permission(i)
 
--- a/x/collection/keeper/supply.go
+++ b/x/collection/keeper/supply.go
@@ -26,15 +26,12 @@ func (k Keeper) CreateContract(ctx sdk.Context, creator sdk.AccAddress, contract
 		Grantee:    creator.String(),
 	}
 	ctx.EventManager().EmitEvent(collection.NewEventGrantPermTokenHead(eventGrant))
-	for _, permission := range collection.Permission_value {
-		p := collection.Permission(permission)
-		if p == collection.PermissionUnspecified {
-			continue
-		}
+	for i := 1; i < len(collection.Permission_value); i++ {
+		p := collection.Permission(i)
 
 		eventGrant.Permission = p
 		ctx.EventManager().EmitEvent(collection.NewEventGrantPermTokenBody(eventGrant))
-		k.Grant(ctx, contractID, []byte{}, creator, collection.Permission(permission))
+		k.Grant(ctx, contractID, []byte{}, creator, p)
 	}
 
 	return contractID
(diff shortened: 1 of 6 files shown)


x/collection/keeper/supply.go
```diff
@@ -26,15 +26,12 @@ func (k Keeper) CreateContract(ctx sdk.Context, creator sdk.AccAddress, contract
...
-		if p == collection.PermissionUnspecified {
-			continue
-		}
+	for i := 1; i < len(collection.Permission_value); i++ {
+		p := collection.Permission(i)
 
 		eventGrant.Permission = p
 		ctx.EventManager().EmitEvent(collection.NewEventGrantPermTokenBody(eventGrant))
-		k.Grant(ctx, contractID, []byte{}, creator, collection.Permission(permission))
+		k.Grant(ctx, contractID, []byte{}, creator, p)
```
zemyblue, line 34: Is this because `collection.Permission_value` is map type?
If then, how can we guarantee that this modification is deterministic?

x/token/module/module.go
```diff
@@ -111,10 +111,14 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	token.RegisterQueryServer(cfg.QueryServer(), keeper.NewQueryServer(am.keeper, am.authKeeper))
 
 	// m := keeper.NewMigrator(am.keeper)
```
jaeseung-bae: How about replacing commented-out code for guidence with `NOTE: Place to add migrator`

x/collection/keeper/supply.go
0Tech: `for i := 1; i < len(collection.Permission_value); i++ {`
guarantees it.

x/token/module/module.go
0Tech: I thought the code would be out-dated, because we won't use the code now. I didn't want to maintain the unused logic, so I removed the entire logics.

jaeseung-bae (approved):

x/collection/keeper/supply.go
```diff
@@ -26,15 +26,12 @@ func (k Keeper) CreateContract(ctx sdk.Context, creator sdk.AccAddress, contract
 		Grantee:    creator.String(),
 	}
 	ctx.EventManager().EmitEvent(collection.NewEventGrantPermTokenHead(eventGrant))
-	for _, permission := range collection.Permission_value {
-		p := collection.Permission(permission)
-		if p == collection.PermissionUnspecified {
-			continue
-		}
+	for i := 1; i < len(collection.Permission_value); i++ {
```
loin3, line 29: It's a little bit reluctant to start from number 1 without any explanation.
Which is better, starting from number 0 as before and continuing in case of number 0 or starting from number 1 as now??

x/collection/keeper/supply.go
zemyblue: Could you add checking deterministic unittest?

x/collection/keeper/supply.go
0Tech: I think it's quite common protobuf use case, but it may be better to add some explanation.
```suggestion
	// 0 is "unspecified"
	for i := 1; i < len(collection.Permission_value); i++ {
```

x/collection/keeper/supply.go
0Tech: We cannot simply check the determinism by checking the apphash after the execution, because, it may not have different root hash (it's the reason why we could not catch this code until now). Any opinions?

0Tech: It might be possible checking the relevant events. But in this case, we should ensure event determinism for the simple unit test code. But the changes would be large, so I cannot simply add the expected modifications in this fix PR.

zemyblue: I thought it would be all right if we had a unittest that would put in a list of multiple permissions ahead of time and make sure that these values always write in the same sequence of events.

jaeseung-bae (approved):

loin3 (approved):

zemyblue (approved):

Issue #962, opened 2023-04-11
Event determinism on x/token and x/collection


0Tech:
It might be possible checking the relevant events. But in this case, we should ensure event determinism for the simple unit test code. But the changes would be large, so I cannot simply add the expected modifications in this fix PR.

_Originally posted by @0Tech in https://github.com/line/lbm-sdk/pull/961#discussion_r1162248655_

PR #963, opened 2023-04-11
fix: check event determinism on original modules

0Tech:
## Description

This patch would make the events deterministic and add the corresponding tests. The target modules are:
- x/foundation
- x/token
- x/collection

closes: #962

## Checklist:


loin3 (dismissed):

zemyblue (dismissed):

zemyblue (approved):

loin3 (approved):

x/collection/keeper/keeper_test.go
```diff
@@ -39,22 +43,30 @@ type KeeperTestSuite struct {
...
-				break
+func (s *KeeperTestSuite) createRandomAccounts(accNum int) []sdk.AccAddress {
+	if s.deterministic {
+		addresses := make([]sdk.AccAddress, accNum)
+		for i := range addresses {
+			addresses[i] = sdk.AccAddress(fmt.Sprintf("address%d", i))
+		}
+		return addresses
+	} else {
+		seenAddresses := make(map[string]bool, accNum)
```
jaeseung-bae, line 54: It's just FYI. You may already know.
Map with an empty struct as a value like this `make(map[string]struct{}, accNum)` doesn't consume any memory.
And We can still check if a key exists in a conventional way `if _, ok := m["key"]; !ok {}`.

jaeseung-bae (approved):

x/collection/keeper/keeper_test.go
0Tech: As you know, it's one of the eternal debates on Go. If the size of the elements is sufficiently large, we should use `struct{}` as you suggested. But in this case, it's just a matter of choice. Thank you for reminding me.

jaeseung-bae: I agree. That works well in case of large one~

PR #966, opened 2023-04-14
fix: remove legacy events on x/collection and x/token

0Tech:
## Description

This patch would remove the legacy events from x/token and x/collection.

## Motivation and context

The legacy API has been already deprecated, and it conflicts with other APIs (e.g. x/bank's).

## Checklist:


x/token/event.go
```diff
@@ -1,28 +1,13 @@
...
 import (
-	"fmt"
-	"sort"
 	"strings"
-
-	sdk "github.com/line/lbm-sdk/types"
 )
 
 const (
-	prefixEventType    = "EVENT_TYPE_"
```
jaeseung-bae, line 12: Is there any side-effect that might happen soon when someone uses the same prefix in the future again?
It's very unlikely It's okay to remove it.

zemyblue (approved):

x/token/event.go
0Tech: The prefix was there just for the legacy event's stringer. We don't support the legacy events anymore, I think it would be OK to remove it. In this PR, the relevant enums would also be removed.

loin3 (approved):

PR #980, opened 2023-04-17
refactor: refactor x/token,collection query errors

0Tech:
## Description

This patch would refactor the query APIs on x/token and x/collection.
The following items are the ground rules over the query APIs.
- If the parameters are invalid regardless of the chain state, it would emit an gRPC error INVALID_ARGUMENT.
- If the response type is of number (e.g. sdk.Int) or boolean, and any of the relevant assets does not exist, it would simply return 0 or false, respectively.
- If the response type is in singular form (e.g. NFT), and any of the relevant assets does not exist, it would emit an gRPC error NOT_FOUND.
- If the response type is in plural form (i.e. list), and any of the relevant assets does not exist, it would simply return empty list.
- The errors caused by the page request would NOT be wrapped by the correct gRPC error, because we cannot simply discriminate it unless we change the upstream's code.

## Motivation and context

The client requested consistent error policy over the query APIs. Also no more backward compatibility is required by the client. So we refactor those APIs before the launch.

## Checklist:


jaeseung-bae (approved):

loin3:
Would you add description when the response type is boolean like hasParent query's response type?

loin3 (approved):

zemyblue (approved):

0Tech:
> Would you add description when the response type is boolean like hasParent query's response type?

I've updated the description. Thank you!

PR #1013, opened 2023-05-19
feat: modify CC, SCC interfaces to process multi-txs

tkxkd0159:
## Description

closes: xxxx

1. All state batches are valid if they have not been deleted during challenge period.
2. The unit of state transition for confirmation is `batch`. If one of intermediate state roots is wrong in specific batch, this batch is rejected and the state root returns to pre-state root.
3. The batch frame matches the L2 block one-on-one, and the L2 batch consists of several batch frames.
4. Queue transaction has three states (pending, finalized, outdated). If this tx is not included in L1 CC batch until deadline, it would be outdated and all registered sequencer get penalty.

## Checklist:


x/or/da/keeper/msg_server.go
```diff
@@ -46,8 +47,11 @@ func (k msgServer) AppendCCBatch(goCtx context.Context, msg *types.MsgAppendCCBa
 		return nil, err
 	}
 
-	k.appendSequencerBatch()
+	ctx := sdktypes.UnwrapSDKContext(goCtx)
 
+	_ = sha256.Sum256(ctx.TxBytes())
+	k.appendSequencerBatch()
+	ctx.HeaderHash()
```
shiki-tak, line 54: I would like to know the reason for this change.

ulbqb:
I think CCBatch interface need L2 block interface. What block is reproduced from CCBatch?

tkxkd0159:
> What kind of blocks are reproduced from CCBatch? I think it's difficult to decide the exact block structure at this stage, but I'd like to know what blocks you have in mind.
> 
> [tendermint Block](https://github.com/tendermint/tendermint/blob/35581cf54ec436b8c37fabb43fdaa3f48339a170/proto/tendermint/types/block.proto#L10-L15) [rollkit Block](https://github.com/rollkit/rollkit/blob/8b1f9b4451d6eb9854b8046edb69e3b2a677d6cc/proto/rollkit/rollkit.proto#L74-L77) [dymint Block](https://github.com/dymensionxyz/dymint/blob/b59ca6f24577101a1ba60f7a16e145c824b1f01e/proto/types/dymint/dymint.proto#L74-L78)

@ulbqb You can reproduce any block with this batch. CC only have duty to save and sequence txs in my design. I assume that any L2 node already know their L2 block scheme So they can reproduce their block with this batch data. When they reconstruct their block, they should fetch their block header from their L2 peer who already have that height block. They can validate this header because the CC batch has parent hash data.

ulbqb:
@tkxkd0159 
My concern is that Batch does not include sequencer signatures. Since the signature can only be created by the sequencer himself, there is a possibility that the L2 block cannot be reconstructed. However, I don't know if the L2 block needs to be signed in the first place, so I think we need to investigate this.

However, I don't think there is a big problem with the structure, so I think the current interface is fine.

ulbqb (approved):

tkxkd0159:
> @tkxkd0159 My concern is that Batch does not include sequencer signatures. Since the signature can only be created by the sequencer himself, there is a possibility that the L2 block cannot be reconstructed. However, I don't know if the L2 block needs signatures in the first place, so I think we need to investigate this.
> 
> However, I don't think there is a big problem with your structure, so I think the current interface is fine.

In Optimism & Arbitrum cases, they assume that only verified sequencer can push DA data to L1. So I guess they guarantee their data integrity without signature. It seems a bit naive and centralized but I think it's fine at our initial L2 layout.

shiki-tak (approved):

PR #1025, opened 2023-06-06
chore: update golang version

shiki-tak:
## Description

This branch uses golang version 1.19 ahead of time.

The fixed test case is similar to cosmos-sdk update from 1.18 to 1.19.

https://github.com/cosmos/cosmos-sdk/compare/v0.45.10...v0.45.16#diff-a0c3450e59ee801d721425680c1a50e836d3e4d7baae7fec2c4ec03ebfdf2d76R82

https://github.com/cosmos/cosmos-sdk/compare/v0.45.10...v0.45.16#diff-1b93526887e605cf730c42e7a98bd245da82183ba2f77536a39d211c4c58ff30R131

## Checklist:


snapshots/manager_test.go
```diff
@@ -79,7 +79,7 @@ func TestManager_Take(t *testing.T) {
 		Height: 5,
 		Format: snapshotter.SnapshotFormat(),
 		Chunks: 1,
-		Hash:   []uint8{0xcd, 0x17, 0x9e, 0x7f, 0x28, 0xb6, 0x82, 0x90, 0xc7, 0x25, 0xf3, 0x42, 0xac, 0x65, 0x73, 0x50, 0xaa, 0xa0, 0x10, 0x5c, 0x40, 0x8c, 0xd5, 0x1, 0xed, 0x82, 0xb5, 0xca, 0x8b, 0xe0, 0x83, 0xa2},
```
torao, line 82: Why does this `Hash` value depend on the go version? Is it mean that snapshots created on go 1.18 nodes and transferred to go 1.19 nodes cannot be restored correctly? Is it the same as the problem for AppHash?

Mdaiki0730:
The following files also mention go version, so why don't you change them?
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/CONTRIBUTING.md?plain=1#L132
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/README.md?plain=1#L7
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/images/simd-dlv/Dockerfile#L1
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/images/simd-env/Dockerfile#L1
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/rosetta/node/Dockerfile#L1
https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/rosetta/rosetta-cli/Dockerfile#L1

snapshots/manager_test.go
shiki-tak: I have even confirmed that this hash is from:
https://github.com/Finschia/finschia-sdk/blob/main/snapshots/store.go#L259

torao: The difference in this `Hash` value between go versions was due to differences in the binaries generated by the compression of the snapshot. By removing the compressed output stream layer from `StreamWriter`, which is responsible to generate the binaries, as follows, it was confirmed that the same binaries are generated and the hash values matched for both 1.18 and 1.19.

StreamWriter with compression process removed

It makes sense for snapshots to be compressed and have a hash value for checksum purposes to detect any loss during transmission. But in this test case, it should be verified that the binary SHA-256 value for the generated snapshot matches the `Hash` value, rather than hard-coding a `Hash` value that depends on the implementation and behaviour of the compression algorithm. This could be done in a separate PR and should be done for the Cosmos SDK.

In this test case, the `Hash` value is found to match the value generated by SHA-256 for concatenated `expectChunks`.

shiki-tak: I also think that it is appropriate to conduct this verification by PR for the main branch of finschia-sdk or the original cosmos-sdk.(Even if you submit a PR to the head office, it may take some time for finschia-sdk to backport the version, so it may be safer to submit a PR to our main branch.)

In any case, in this PR, I think that the hash value change due to version update of go can be left as it is.

shiki-tak:
> The following files also mention go version, so why don't you update them? https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/CONTRIBUTING.md?plain=1#L132 https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/README.md?plain=1#L7 https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/images/simd-dlv/Dockerfile#L1 https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/images/simd-env/Dockerfile#L1 https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/rosetta/node/Dockerfile#L1 https://github.com/shiki-tak/finschia-sdk/blob/update-golang-version/contrib/rosetta/rosetta-cli/Dockerfile#L1

It seems that there was a part that could not be replaced. As for the document, the head family also updated it in the same way, so I updated it here as well.

Mdaiki0730 (approved):
LGTM

snapshots/manager_test.go
torao: The relevant fixes aren't the purpose of this PR and should be a separate PR.

The root cause of the different compression binary is explained in the following cosmos-sdk issue: before 1.18, the `Close()` is called twice to add two trailers, and after 1.19 the second `Close()` is ignored and only one trailer is added.
https://github.com/cosmos/cosmos-sdk/issues/13373

PR #1078, opened 2023-08-10
chore: fix tag comments in github actions workflow docker.yml

loloicci:
This PR fixes some comments in workflow.
This workflow does not triggered tags without patch number.

## Checklist:

- [ ] I have added tests to cover my changes. (not needed)
- [ ] I have updated the documentation accordingly. (not needed)
- [ ] I have updated API documentation `client/docs/swagger-ui/swagger.yaml`(not needed)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,8 +6,8 @@ on:
       - rc*/*
   push:
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
-      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v1.0-rc1, v20.15.10-rc5
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v20.15.10-rc5, v20.15.10-rc12+build1
 env:
   registry: docker.io
   repository: finschia/finschia-simapp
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -6,8 +6,8 @@ on:
       - rc*/*
   push:
     tags:
-      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v1.0, v20.15.10
-      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v1.0-rc1, v20.15.10-rc5
+      - "v[0-9]+.[0-9]+.[0-9]+" # Push events to matching v*, i.e. v20.15.10
+      - "v[0-9]+.[0-9]+.[0-9]+-rc*" # Push events to matching v*, i.e. v20.15.10-rc5, v20.15.10-rc12+build1
```
zemyblue, line 10: Is It possible using `+` as tag in golang. 
Because I'm not sure it is no problem when other service import the tag with `+`.

loloicci: I don't know that, but this pattern matches with `v20.15.10-rc12+build1`. If it cannot be used as version, this pattern should be fixed.

loloicci: You can use `+` in tag. This can be confirmed with https://github.com/loloicci/gomodexample/tags

zemyblue (approved):

.github/workflows/docker.yml
jaeseung-bae, line 10:
```suggestion
      - "v[0-9]+.[0-9]+.[0-9]+-rc.*" # Push events to matching v*, i.e. v20.15.10-rc5, v20.15.10-rc12+build1
```
Don't we need `.` before `*` in order to include the words(5, 12+build1) right after `rc`?
When I try this in online regex site(https://regex101.com/), This doesn't seem to match to `v20.15.10-rc5, v20.15.10-rc12+build1`
Could you check it?

loloicci: This is not regex and this pattern is according to https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

jaeseung-bae: Okay. Thank you for the link.

jaeseung-bae (approved):

PR #1081, opened 2023-08-17
docs: document default events

0Tech:
## Description

This patch would add documentation about default events by baseapp.

refer: #1063

## Checklist:

- [x] I followed the [contributing guidelines](https://github.com/Finschia/finschia-sdk/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/Finschia/finschia-sdk/blob/main/CODE_OF_CONDUCT.md).
- [x] I have added a relevant changelog to `CHANGELOG.md`
- [ ] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated API documentation `client/docs/swagger-ui/swagger.yaml`

PR #1083, opened 2023-08-17
docs: add detailed explanation about default events

ryu1-sakai:
## Description

It seems better to add more detailed information about default events emitted by baseapp. (Especially information about the event structure.)

This PR is trying to add such information to the document created by #1081.

## Checklist:

### DIFF
--- a/docs/docs/core/08-events.md
+++ b/docs/docs/core/08-events.md
@@ -11,9 +11,15 @@ sidebar_position: 1
 
 There are a few events that are automatically emitted for all messages, directly from `baseapp`.
 
-* `message.action`: The name of the message type. Exactly one exists for each transaction message, in the first ABCI `Event` of each transaction message.
-* `message.sender`: The address of the message signer. Exactly one exists for each transaction message, in the first ABCI `Event` of each transaction message.
+* `message.action`: The name of the message type.
+* `message.sender`: The address of the message signer.
 * `message.module`: The name of the module that emitted the message.
+
+:::tip
+`baseapp` emits exactly one `message` event for each message before any other events emitted by the message.
+The `message` event contains at least 2 attributes, exactly one `action` and exactly one `sender`.
+The position of the event may change in the next major version.
+:::
 
 :::tip
 The module name is assumed by `baseapp` to be the second element of the message route: `"cosmos.bank.v1beta1.MsgSend" -> "bank"`.
--- a/docs/docs/core/08-events.md
+++ b/docs/docs/core/08-events.md
@@ -11,8 +11,8 @@ sidebar_position: 1
 
 There are a few events that are automatically emitted for all messages, directly from `baseapp`.
 
-* `message.action`: The name of the message type.
-* `message.sender`: The address of the message signer.
+* `message.action`: The name of the message type. Exactly one exists for each transaction message, in the first ABCI `Event` of each transaction message.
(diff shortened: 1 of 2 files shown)


docs/docs/core/08-events.md
```diff
@@ -11,8 +11,8 @@ sidebar_position: 1
 
 There are a few events that are automatically emitted for all messages, directly from `baseapp`.
 
-* `message.action`: The name of the message type.
-* `message.sender`: The address of the message signer.
+* `message.action`: The name of the message type. Exactly one exists for each transaction message, in the first ABCI `Event` of each transaction message.
+* `message.sender`: The address of the message signer. Exactly one exists for each transaction message, in the first ABCI `Event` of each transaction message.
```
zemyblue, line 15: I think each `in the first ABCI Event of each transaction message.` sentence seems to be a little conflict. Because if all attributes are existed, it can confuse which one should be first according to this sentence. It is a little my opinion. 

And how about moving the additional description as a tip. I think it's nice to explain the default information about each event attribute, and explain more information as a tip.

docs/docs/core/08-events.md
```diff
@@ -11,8 +11,8 @@ sidebar_position: 1
...
 * `message.module`: The name of the module that emitted the message.
 
 :::tip
```
0Tech, line 18: If you don't mind:

```suggestion
* `message.action`: The name of the message type.
* `message.sender`: The address of the message signer.
* `message.module`: The name of the module that emitted the message.

:::tip
`baseapp` emits exactly one `message` event before any other events emitted by the message, which contains two attributes at least, `action` and `sender`. The position of the event may change in the next major version.
```

ryu1-sakai: @zemyblue and @0Tech, thank you for your comments.

I agree with you, and tried to improve an unclear point in @0Tech's suggestion. How about the following?

```
## Default Events

There are a few events that are automatically emitted for all messages, directly from `baseapp`.

* `message.action`: The name of the message type.
* `message.sender`: The address of the message signer.
* `message.module`: The name of the module that emitted the message.

:::tip
`baseapp` emits exactly one `message` event for each message before any other events emitted by the message.
The event contains at least 2 attributes, exactly one `action` and exactly one`sender`.
The position of the event may change in the next major version.
:::

:::tip
The module name is assumed by `baseapp` to be the second element of the message route: `"cosmos.bank.v1beta1.MsgSend" -> "bank"`.
In case a module does not follow the standard message path, (e.g. IBC), it is advised to keep emitting the module name event.
`Baseapp` only emits that event if the module have not already done so.
:::
```

zemyblue: How about adding more detain about event like bellow. 
```
The `message` event contains at least 2 attributes, exactly one `action` and exactly one `sender`.
```
or
```
The `message` event contains at least two properties: 'action' and 'sender'.
```

ryu1-sakai: Thank you for the suggestion. I like the first one. It gives more information than the latter one.

```
The `message` event contains at least 2 attributes, exactly one `action` and exactly one `sender`.
```

zemyblue: Please update document.

ryu1-sakai:
@zemyblue, PTAL

zemyblue (commented):
@ryu1-sakai , please resolve conflicts

ryu1-sakai:
@zemyblue Rebased.

zemyblue (approved):

tkxkd0159 (approved):

Issue #1084, opened 2023-08-18
Support compatibility with the Cosmos Query APIs


zemyblue:
## Summary

Support compatibility with the Cosmos Query APIs.

**proto/cosmos/base/tendermint/v1beta1/query.proto**
* `/cosmos/base/tendermint/v1beta1/node_info`
* `/cosmos/base/tendermint/v1beta1/syncing`
* `/cosmos/base/tendermint/v1beta1/blocks/latest`
* `/cosmos/base/tendermint/v1beta1/blocks/{height}`
* `/cosmos/base/tendermint/v1beta1/validatorsets/latest`
* `/cosmos/base/tendermint/v1beta1/validatorsets/{height}`

**proto/cosmos/tx/v1beta1/service.proto**
* `/cosmos/tx/v1beta1/txs`

## Problem Definition

The some query apis are not supported in finschia. So we cannot use any cosmos infra without modification because it makes a compatibility problem with cosmos. The basic problem is difference the block type between tendermint and ostracon. A entropy data was added more in ostracon. So I think we can convert it for tendermint. 

## Proposal

____

#### For Admin Use

- [x] Not duplicate issue
- [ ] Appropriate labels applied
- [ ] Appropriate contributors tagged
- [ ] Contributor assigned/self-assigned

PR #982, opened 2023-04-18
fix: not to throw error when no txs in block

loin3:
## Description

I followd this [cosmos-sdk/issues/12040](https://github.com/cosmos/cosmos-sdk/issues/12040) and [cosmos-sdk/pull/12108](https://github.com/cosmos/cosmos-sdk/pull/12108)

our version 0.45.x does not have `GetBlockWithTxs` in `x/auth/tx/service`, so we create `x/auth/tx2`.
It's hard to cherr-pick because the directories are different. So I reflected the changes directly into tx2.

## Motivation and context

When query through `GetBlockWithTxs`, `invalid request` error occurred when no txs are included in block.
The correct behavior is to return the block with empty list of txs.

## How has this been tested?

tested no error occured and return 0 length of tx list

## Screenshots (if appropriate):
before thsi PR, when tx does not exist in block
```
{
  "code": 3,
  "message": "out of range: cannot paginate 0 txs with offset 0 and limit 100: invalid request: invalid request",
  "details": []
}
```

when tx does not exist in block
```
{
  "txs": [],
  "block_id": {
    "hash": "tMvoEOdixC8Kt9YNwb6vGDsYJO/euSW3TRuDfzLijPk=",
    "part_set_header": {
      "total": 1,
      "hash": "ykh4pBbXq5ehsj2zyFRv0UdOFnf/8rVwjVl5MXN+pFM="
    }
  },
  "block": {
    "header": {
      "version": {
        "block": "11",
        "app": "0"
      },
      "chain_id": "simd-testing",
      "height": "115",
      "time": "2023-04-18T08:51:57.136998Z",
      "last_block_id": {
        "hash": "2mVthS8/78uoXxmNaQaO43UivRU142JD42xE/d3ps4I=",
        "part_set_header": {
          "total": 1,
          "hash": "ucOMMUTAhQtMR8sREYbbyG6RKKIDb3bOxjh5D/j1s0M="
        }
      },
      "last_commit_hash": "sUmrUkJM+xlOfckEN7yXVJX9zFySepZa2Gnl670tzwQ=",
      "data_hash": "47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=",
      "validators_hash": "wUFwitwdiD2JQIkDe4MofWg7YqGlxwofCZBWzI7H5Mk=",
      "next_validators_hash": "wUFwitwdiD2JQIkDe4MofWg7YqGlxwofCZBWzI7H5Mk=",
[...]

PR #891, opened 2023-02-10
feat!: add custom `GetBlockWithTxs` for lbm

zemyblue:
### move `GetBlockWithTxs` of lbm
* roll back origin `cosmos.tx.v1beta1.Service/GetBlockWithTxs`
* add `lbm.tx.v1beta1.Service/GetBlockWithTxs`.
* `lbm.tx.v1beta1.service.proto` converted in `x/auth/tx2` directory
* `cosmos.tx.v1beta1.Service/GetBlockWithTxs` query api is deprecated.

related: #885 

## Checklist:

- [x] I followed the [contributing guidelines](https://github.com/line/lbm-sdk/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/line/lbm-sdk/blob/main/CODE_OF_CONDUCT.md).
- [x] I have added a relevant changelog to `CHANGELOG.md`
- [x] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [x] I have updated API documentation `client/docs/swagger-ui/swagger.yaml`

PR #1085, opened 2023-08-18
feat: rollback `GetBlockWithTxs` of cosmos service.proto for compatibility with cosmos-sdk APIs

zemyblue:
## Description

related: #1084 

Rollback the function not supported `cosmos.tx.v1beta1.Service/GetBlockWithTxs` api of cosmos service.proto by #891 and apply the bug fix of #982 .

## Checklist:


jaeseung-bae (dismissed):

x/auth/tx/service.go
```diff
@@ -161,9 +164,90 @@ func (s txServer) GetTx(ctx context.Context, req *txtypes.GetTxRequest) (*txtype
...
+		offset = req.Pagination.Offset
+		limit = req.Pagination.Limit
+	} else {
+		offset = 0
+		limit = pagination.DefaultLimit
+	}
+
+	blockTxs := block.Data.Txs
+	blockTxsLn := uint64(len(blockTxs))
+	txs := make([]*txtypes.Tx, 0, limit)
```
jaeseung-bae, line 204: `txs := make([]*txtypes.Tx, 0, limit)`
I wonder what's happening when the limit value is the max value of uint64.
Can it be an attack point to make the chain allocate lots of memory unnecessarily?

0Tech (commented):
This api also returns `Block` which lacks some information in the original one. It would be better to add comments, as #1087 did.

x/auth/tx/service.go
zemyblue: Yes, it can be. And we proposed this problem to upstream (cosmos-sdk/issues/14917). So we will apply the changes later.

jaeseung-bae (dismissed):

0Tech (approved):

jaeseung-bae (approved):

PR #1086, opened 2023-08-18
fix: fix redundant priv-validator-key generation when running with kms

jaeseung-bae:
## Description

* Fix for redundant key generation. With running kms, generating priv-key is unnecessary.

## Checklist:

### DIFF
--- a/server/start_test.go
+++ b/server/start_test.go
@@ -9,10 +9,13 @@ import (
 
 func TestGenPvFileOnlyWhenKmsAddressEmptyGenerateFiles(t *testing.T) {
 	cfg := config.DefaultConfig()
-	cfg.PrivValidatorKey = "./key.json"
-	cfg.PrivValidatorState = "./state.json"
-	defer os.Remove(cfg.PrivValidatorKey)
-	defer os.Remove(cfg.PrivValidatorState)
+	dir, err := os.MkdirTemp("", "start_test")
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+	defer os.RemoveAll(dir)
+	cfg.PrivValidatorKey = dir + "/key.json"
+	cfg.PrivValidatorState = dir + "/state.json"
 
 	pv := genPvFileOnlyWhenKmsAddressEmpty(cfg)
 
--- a/server/start_test.go
+++ b/server/start_test.go
@@ -1,0 +1,29 @@
+package server
+
+import (
+	"github.com/Finschia/ostracon/config"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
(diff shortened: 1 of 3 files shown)


.github/workflows/check-generated.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          make error-doc-gen
+          if ! git diff --stat --exit-code ; then
+            echo ">> ERROR:"
+            echo ">>"
+            echo ">> Error documents require update (source files in x folder may have changed)."
+            echo ">> Ensure your tools are up-to-date, re-run 'make error-doc' and update this PR."
+            echo ">>"
+            exit 1
+          fi
```
zemyblue, line 37: Why does this file is added in this file?

jaeseung-bae: This was added automatically when I click github ui to change base. I'll fix it.

zemyblue (commented):
I think it's nice to add some explanation about doesn't generate a private key if some condition.

jaeseung-bae:
> I think it's nice to add some explanation about doesn't generate a private key if some condition.

I named the function `genPvFileOnlyWhenKmsAddressEmpty` to express conditional generation of the key. 
The name doesn't seem to be enough. 
Do you have any idea?

jaeseung-bae:
> > > I think it's nice to add some explanation about doesn't generate a private key if some condition.
> > 
> > 
> > I named the function `genPvFileOnlyWhenKmsAddressEmpty` to express conditional generation of the key. The name doesn't seem to be enough. Do you have any idea?
> 
> I think the condition of which does not generate a private key seems that the config.yml set `priv_validator_laddr` is empty string and there is no private key. Is it right?
> 
> If then, how about adding this explanation in config.yml. Oh, it is Ostracon's area. right?

If `priv_validator_laddr` is empty string, no key will be generated.
It's nice to put an explanation to config.yml. 
Yes, It's ostracon's area. I'll put it in ostracon's area.

zemyblue (dismissed):

server/start_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/Finschia/ostracon/config"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestGenPvFileOnlyWhenKmsAddressEmptyGenerateFiles(t *testing.T) {
+	cfg := config.DefaultConfig()
+	cfg.PrivValidatorKey = "./key.json"
```
0Tech, line 12: How about using `os.MkdirTemp()`?

0Tech (approved):

zemyblue (approved):

PR #1105, opened 2023-08-31
fix: add minted coins to balance in x/collection MsgMintFT

0Tech:
## Description

This patch would fix accumulation logic of balance in MsgMintFT.

closes: #1104

## Checklist:

- [x] I followed the [contributing guidelines](https://github.com/Finschia/finschia-sdk/blob/main/CONTRIBUTING.md) and [code of conduct](https://github.com/Finschia/finschia-sdk/blob/main/CODE_OF_CONDUCT.md).
- [x] I have added a relevant changelog to `CHANGELOG.md`
- [x] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated API documentation `client/docs/swagger-ui/swagger.yaml`

PR #1106, opened 2023-09-01
fix: support x/collection migration on old chains

0Tech:
## Description

This patch would support old chains in which the default class ids start from 0.

related: #1105

## Checklist:


zemyblue (commented):
Could you add any unittest for this change?

0Tech:
> Could you add any unittest for this change?

Instead of adding a unit test, I've added a comment explaining the reason of the change.

jaeseung-bae (approved):

PR #1133, opened 2023-10-04
test(collection): add more verify logics for tests(SendFT, OperatorSendFT, Authorize, Revoke)

jaeseung-bae:
## Description

* This PR refactors tests for `SendFT`, `OperatorSendFT`, `AuthorizeOperator`, and `RevokeOperator` to check more states

## Checklist:

### DIFF
--- a/x/collection/keeper/msg_server_test.go
+++ b/x/collection/keeper/msg_server_test.go
@@ -220,16 +220,18 @@ func TestHelperBuildEventSent(t *testing.T) {
 			amount:     collection.NewCoins(collection.NewFTCoin("00000001", sdk.NewInt(1000000))),
 		},
 	}
-	for _, tc := range testCases {
-		// Act
-		from, err := sdk.AccAddressFromBech32(tc.from)
-		assert.NoError(t, err)
-		to, err := sdk.AccAddressFromBech32(tc.to)
-		assert.NoError(t, err)
-		event := helperBuildEventSent(tc.contractID, from, to, from, tc.amount)
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			// Act
+			from, err := sdk.AccAddressFromBech32(tc.from)
+			assert.NoError(t, err)
+			to, err := sdk.AccAddressFromBech32(tc.to)
+			assert.NoError(t, err)
+			event := helperBuildEventSent(tc.contractID, from, to, from, tc.amount)
 
-		// Assert
-		assert.Equal(t, tc.expectedEvent, event)
+			// Assert
+			assert.Equal(t, tc.expectedEvent, event)
+		})
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zemyblue (commented):
Are there any other more test cases according decision table?

jaeseung-bae:
> Are there any other more test cases according decision table?

No more tests found.

zemyblue (approved):

x/collection/keeper/msg_server_test.go
```diff
@@ -12,122 +14,190 @@ import (
...
+				Amount:     collection.NewCoins(collection.NewFTCoin(s.ftClassID, s.balance.Add(sdk.OneInt()))),
+			},
+			ftID:          collection.NewFTID(s.ftClassID),
+			expectedError: collection.ErrInsufficientToken,
 		},
 	}
 
 	for name, tc := range testCases {
 		s.Run(name, func() {
-			ctx, _ := s.ctx.CacheContext()
```
0Tech, line 39: Is it OK to stop using the cache? Generally, it will make other tests affected by it, which often means, making a race condition. Of course, in this case, it may not be the case.

x/collection/keeper/msg_server_test.go
```diff
@@ -268,109 +338,155 @@ func (s *KeeperTestSuite) TestMsgOperatorSendNFT() {
...
+			s.Require().Equal(tc.req.Holder, prevAuth.Holder)
+			s.Require().Equal(tc.req.Operator, prevAuth.Operator)
+			curAuth, err := s.keeper.GetAuthorization(s.ctx, tc.req.ContractId, holder, operator)
+			s.Require().ErrorIs(err, collection.ErrCollectionNotApproved)
+			s.Require().Nil(curAuth)
 		})
 	}
 }
 
+func helperBuildEventAuthRelated(evtType, contractID string, holder, operator sdk.AccAddress) sdk.Event {
```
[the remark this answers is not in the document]
jaeseung-bae: To prevent event-breaking, I should add unit test for `helperBuildEventAuthRelated` with the original literal events to verify event-breaking. 
Do I understand the point? If so, I'll add unit test for it.

x/collection/keeper/msg_server_test.go
jaeseung-bae: I'll look at it again.

x/collection/keeper/msg_server_test.go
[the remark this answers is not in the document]
0Tech: > I should add unit test for `helperBuildEventAuthRelated` with the original literal events to verify event-breaking.

:+1: And also for other helpers.

I forgot to mention about the other events. It seems like the new tests check only the last event. So you need to add a line to check the last event is the only event caused by the request, hence there is no other event.

x/collection/keeper/msg_server_test.go
```diff
@@ -198,6 +194,36 @@ func helperBuildEventSent(contractID string, from, to, operator sdk.AccAddress,
...
+			contractID:    s.contractID,
+			from:          s.vendor,
+			to:            s.customer,
+			operator:      s.vendor,
+			amount:        collection.NewCoins(collection.NewFTCoin(s.ftClassID, s.balance)),
+		},
+	}
+	if !s.deterministic {
```
0Tech, line 215: Or, you may create your own literals to remove this conditional.

jaeseung-bae: I tried but I couldn't.
I need a different literal for `s.deterministic` because of the difference of the addresses per `s.deterministic`.
Do you have any idea that I could remove this conditional?

0Tech: You should create new addresses in this test. Don't try to use the addresses made by the suite.

jaeseung-bae: Okay, I may add some literal input for the event that can be used regardless of changes.

jaeseung-bae: Okay, I may add some literal input for the event that can be used regardless of changes.

x/collection/keeper/msg_server_test.go
```diff
@@ -515,35 +524,44 @@ func helperBuildEventAuthRelated(evtType, contractID string, holder, operator sd
...
+			}},
+			contractID: "9be17165",
+			holder:     "link1v9jxgun9wdenyjqyyxu",
+			operator:   "link1v9jxgun9wdenzw08p6t",
 		},
 	}
-	if !s.deterministic {
-		return
-	}
 	for _, tc := range testCases {
```
0Tech, line 555: It would be nicer to have "name"s on each case.

0Tech (approved):

zemyblue (approved):

PR #1140, opened 2023-10-16
test(token): refactoring testcase for token

jaeseung-bae:
## Description

* Just refactoring testcases for token module

## Checklist:

### DIFF
--- a/testutil/encoding.go
+++ b/testutil/encoding.go
@@ -16,10 +16,14 @@ func MustJSONMarshal(v any) []byte {
 	return b
 }
 
-type FmtStringer interface {
-	string | sdk.AccAddress | sdk.Int
+// W wraps input with double quotes if it is a string or fmt.Stringer.
+func W(input any) []byte {
+	switch input.(type) {
+	case string, fmt.Stringer:
+		return []byte(fmt.Sprintf("\"%s\"", input))
+	default:
+		panic("unsupported type")
+	}
 }
-
-func W[T FmtStringer](input T) []byte {
 	return []byte(fmt.Sprintf("\"%s\"", input))
 }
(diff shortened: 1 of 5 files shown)


zemyblue (approved):

testutil/encoding.go
```diff
@@ -14,6 +16,10 @@ func MustJSONMarshal(v any) []byte {
 	return b
 }
 
-func W(input string) []byte {
+type FmtStringer interface {
+	string | sdk.AccAddress | sdk.Int
+}
+
+func W[T FmtStringer](input T) []byte {
```
tkxkd0159, line 23: If you want general wrapping, what about use `fmt.Stringer` interface instead of custom generic?

jaeseung-bae: I tried with `fmt.Stringer`at the first time.
But It didn't work for `string` which was a little bit surprising.
Someone said that Builtin types doesn't implement any interfaces(https://groups.google.com/g/golang-nuts/c/hLuHW-ApzH8/m/7tMedjvP6WwJ)

tkxkd0159: Then what about this?

```suggestion
// W wraps input with double quotes if it is a string or fmt.Stringer.
func W(input any) []byte {
	switch input.(type) {
	case string, fmt.Stringer:
		return []byte(fmt.Sprintf("\"%s\"", input))
	default:
		panic("unsupported type")
	}
}
```

zemyblue: But I think it's not good. because this change has error case and we can know it when run time. But in general type, we can know the error when compile time.

jaeseung-bae: Thanks for pointing out the pros and cons.
The function expects any argument that can be stringified. 
New comment seems to explain its purpose better.

Although it checks types at runtime, it doesn't need to add a new type to `string | sdk.AccAddress | sdk.Int` which is good point.

If you don't mind, I would keep it.

tkxkd0159 (approved):
LGTM

zemyblue (approved):

PR #1225, opened 2024-02-02
feat: update x/stakingplus to use Finschia/cosmos-sdk

ulbqb:
## Description

closes: #1200

## Checklist:


x/stakingplus/keeper/keeper_test.go
```diff
@@ -0,0 +1,133 @@
...
+		return false
+	}
+
+	resp, err := c.authz.Accept(sdk.Context{}, msg)
+	return resp.Accept && (err == nil)
+}
+
+func (c CreateValidatorAuthorizationMatcher) String() string {
+	return fmt.Sprintf("grants %s to %s", c.authz.MsgTypeURL(), c.authz.ValidatorAddress)
+}
```
tkxkd0159, line 133: We don't need this part for mock testing based on previous comment. We can assume our authorization will be registered by foundation module properly.

x/stakingplus/keeper/keeper_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// approve Msg/CreateValidator to grantee
+	foundationKeeper.
+		EXPECT().
+		Accept(gomock.Any(), s.grantee, NewCreateValidatorAuthorizationMatcher(s.grantee)).
+		Return(nil)
+	foundationKeeper.
+		EXPECT().
+		Accept(gomock.Any(), gomock.Any(), gomock.Any()).
+		Return(sdkerrors.ErrUnauthorized)
+}
```
tkxkd0159, line 103: The expectations should be different for every test cases. So we shouldn't use this logic in `SetupTest()`. Because it's run every unit test in this suite. I recommend this logic to specific test function. And `CreateValidatorAuthorizationMatcher` doesn't match with real parameter, `sdk.Msg`. So it probably confuses people. I think it's better to implement the test from behavior perspective. We can verify `CreateValidatorAuthorization` at E2E test.

tkxkd0159 (changes requested):

tkxkd0159:
I have two more questions.
1. Is there any reason to not implement e2e test for the special case of stakingplus (createValidator)?
2. Is there any reason to not apply this at app_v2?

ulbqb:
@tkxkd0159 

> Is there any reason to not implement e2e test for the special case of stakingplus (createValidator)?

This time I simply converted the existing stakingplus module. But it seems necessary so I'll add the test.

> Is there any reason to not apply this at app_v2?

https://github.com/ulbqb/finschia-sdk/blob/dcdcfa204220529bc1f308f56523b967edfcacc1/simapp/app_config.go#L78
The initialization of the stakingplus keeper is done here. The keeper interface is not changed so there is few changes.

x/stakingplus/keeper/keeper_test.go
ulbqb: https://github.com/golang/mock/blob/674a1c4a938e7b453416052327872571656801b7/gomock/matchers.go#L23-L31
Matcher is defined by GoMock. This is an abstract representation of input values.

ulbqb: I'll move the foundationKeeper mock to each test.

tkxkd0159: I meant we should populate gomock.Eq(MsgCreateValidator) rather than implementing separate matcher. For example,  we can handle this in TestMsgCreateValidator with initializing gomock EXPECT. We expect that correct authorization will be returned depending on MsgTypeURL in real situation.

x/stakingplus/authz.go
```diff
@@ -0,0 +1,36 @@
...
+
+	if mCreate.ValidatorAddress != a.ValidatorAddress {
+		return foundation.AcceptResponse{}, sdkerrors.ErrUnauthorized.Wrap("validator address differs from the authorization's")
+	}
+
+	return foundation.AcceptResponse{Accept: true}, nil
+}
+
+func (a CreateValidatorAuthorization) ValidateBasic() error {
+	if _, err := sdk.ValAddressFromBech32(a.ValidatorAddress); err != nil {
```
0Tech, line 31: We should not use global codec.

0Tech (changes requested):

x/stakingplus/authz.go
ulbqb: Is it ok to always return nil because this can't use keeper codec?

x/stakingplus/keeper/msg_server_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	for name, tc := range testCases {
+		s.Run(name, func() {
+			ctx, _ := s.ctx.CacheContext()
+
+			pk := simtestutil.CreateTestPubKeys(1)[0]
+			delegation := sdk.NewCoin(sdk.DefaultBondDenom, math.OneInt())
+			req, err := stakingtypes.NewMsgCreateValidator(
+				sdk.ValAddress(tc.delegator).String(),
```
0Tech, line 55: Please mind the codec.

x/stakingplus/module/module.go
```diff
@@ -0,0 +1,185 @@
...
+	if in.Config.Authority != "" {
+		authority = authtypes.NewModuleAddressOrBech32Address(in.Config.Authority)
+	}
+
+	k := stakingkeeper.NewKeeper(
+		in.Cdc,
+		in.StoreService,
+		in.AccountKeeper,
+		in.BankKeeper,
+		authority.String(),
```
0Tech, line 179: Using the global codec.

x/stakingplus/keeper/keeper_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// the work around would be registering a new module account on this suite
+		// because x/bank already has dependency on x/mint, and we must have dependency
+		// on x/bank, it's OK to use x/mint here.
+		minterName := minttypes.ModuleName
+		err := s.bankKeeper.MintCoins(s.ctx, minterName, amount)
+		s.Require().NoError(err)
+
+		minter := s.accountKeeper.GetModuleAccount(s.ctx, minterName).GetAddress()
+		err = s.bankKeeper.SendCoins(s.ctx, minter, holder, amount)
+		s.Require().NoError(err)
```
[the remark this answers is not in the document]
ulbqb: It handles Stake-msgServer as an interface and removes dependencies.

x/stakingplus/README.md
```diff
@@ -0,0 +1,32 @@
...
+A validator is created using the `Msg/CreateValidator` service message.
+
++++ https://github.com/cosmos/cosmos-sdk/blob/v0.50.2/proto/cosmos/staking/v1beta1/tx.proto#L20-L21
+
++++ https://github.com/cosmos/cosmos-sdk/blob/v0.50.2/proto/cosmos/staking/v1beta1/tx.proto#L50-L73
+
+This service message is expected to fail if:
+
+- one of the conditions described in the staking module of the Cosmos-SDK is met.
+- the operator address is not registered on x/foundation through [MsgGrant](https://github.com/Finschia/finschia-sdk/tree/main/x/foundation#msggrant) with `CreateValidatorAuthorization`.
```
0Tech, line 30:
```suggestion
- the operator address is not registered on x/foundation through [MsgGrant](https://github.com/Finschia/finschia-sdk/tree/main/x/foundation/README.md#msggrant) with `CreateValidatorAuthorization`. 
```

0Tech (dismissed):
LGTM

0Tech (dismissed):

jaeseung-bae (approved):
LGTM

x/stakingplus/keeper/msg_server_test.go
```diff
@@ -0,0 +1,73 @@
...
+			valid:     false,
+			acceptRet: sdkerrors.ErrUnauthorized,
+		},
+	}
+
+	for name, tc := range testCases {
+		s.Run(name, func() {
+			s.foundationKeeper.
+				EXPECT().
+				Accept(gomock.Any(), gomock.Any(), gomock.Any()).
```
tkxkd0159, line 41: You can make more sophisticated test if you use gomock.Eq(dummyGrantee), gomock.Eq(req).
req is the msg you generate below. you can use dummy msg for failed case. But it's trivial advice so you can merge.

tkxkd0159 (approved):

0Tech (approved):

jaeseung-bae (approved):

PR #1239, opened 2024-02-14
feat: add `CalculateAllInputsHash` in ZKAuthInputs

zemyblue:
## Description

Add `CalculateAllInputsHash` of ZKAuthInputs in x/zkauth module

## Checklist:


x/zkauth/types/zkauth_inputs.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return hash, nil
+}
+
+// CircomBigIntToChunkedBytes converts a big integer to a slice of chunked *big.Int.
+func CircomBigIntToChunkedBytes(num *big.Int) []*big.Int {
+	bytesPerChunk, numChunks := CircomBigintN, CircomBigintK
```
Mdaiki0730, line 112: Checking the movement, it seems that this is actually bitsPerChunk.

x/zkauth/types/zkauth_inputs.go
```diff
@@ -0,0 +1,199 @@
...
+	strPadded := make([]byte, maxSize)
+	copy(strPadded, bytes)
+
+	const chunkSize = PackWidth / 8
+	chunks := ChunkArray(strPadded, chunkSize)
+	packed := make([]*big.Int, 0, len(chunks))
+	for _, chunk := range chunks {
+		byteChunk := make([]byte, len(chunk))
+		copy(byteChunk, chunk)
+		packed = append(packed, BytesBEToBigInt(byteChunk))
```
Mdaiki0730, line 99: Why use a copy of the chunk? I think you can use it directly.

x/zkauth/types/zkauth_inputs.go
```diff
@@ -0,0 +1,199 @@
...
+			end = len(revArray)
+		}
+		chunk := make([]byte, end-start)
+		copy(chunk, revArray[start:end])
+		reverse(chunk)
+		chunks[i] = chunk
+	}
+
+	// return the final chunk array reverted.
+	reverse(chunks)
```
Mdaiki0730, line 69: By specifying the array range, it can be written more simply than in Javascript. The amount of calculation will also be reduced from O(n^2) to O(n) since reverse is eliminated.
```suggestion
	shouldFirstPack := chunkSize
	if len(array)%chunkSize != 0 {
		shouldFirstPack = len(array) % chunkSize
	}
	for i := 0; i < chunkCount; i++ {
		if i == 0 {
			chunks[i] = array[0:shouldFirstPack]
		} else {
			chunks[i] = array[shouldFirstPack+((i-1)*chunkSize) : shouldFirstPack+(i*chunkSize)]
		}
	}
```

x/zkauth/types/zkauth_inputs.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, sdkerrors.Wrap(ErrInvalidZkAuthInputs, "invalid address_seed")
+	}
+	ephPkFrs := SplitToTwoFrs(ephPkBytes)
+	maxBlockHeightFr := new(big.Int).SetInt64(maxBlockHeight)
+
+	issBytes, err := base64.StdEncoding.DecodeString(zk.IssF)
+	if err != nil {
+		return nil, err
+	}
+	issBase64Fr, err := HashASCIIStrToField(string(issBytes), MaxExtIssLen)
```
Mdaiki0730, line 173: Isn't it issFr because the decoded iss is HashASCIIStrToField?

Mdaiki0730:
The hashed element for poseidon hash is written as Fr, but there are some that are already written as F (e.g. IssF), which can cause confusion.
There are many things that already use F, so why not write it as F?

x/zkauth/types/zkauth_inputs.go
zemyblue: Sorry, `bytesPerChunk` is right.

x/zkauth/types/zkauth_inputs.go
zemyblue: Good point. 
This code is legacy before solving revert array.
Thank you, I'll remove it.

zemyblue:
> The hashed element for poseidon hash is written as Fr, but there are some that are already written as F (e.g. IssF), which can cause confusion. There are many things that already use F, so why not write it as F?

I use gnark style, but it's better to use `f`. I'll modify it. Thank you.

x/zkauth/types/zkauth_inputs.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		return poseidon.Hash([]*big.Int{hash1, hash2})
+	default:
+		return nil, errors.Errorf("Yet to implement: unable to hash a vector of length %d", arrayCnt)
+	}
+}
+
+func (zk *ZKAuthInputs) CalculateAllInputsHash(ephPkBytes, modulus []byte, maxBlockHeight int64) (*big.Int, error) {
+	if utf8.RuneCountInString(zk.HeaderBase64) > MaxHeaderLen {
+		return nil, sdkerrors.Wrap(ErrInvalidZkAuthInputs, "zkAuth header should be less then MAX_HEADER_LEN")
```
shiki-tak, line 143:
```suggestion
		return nil, sdkerrors.Wrap(ErrInvalidZkAuthInputs, "zkAuth header should be less than MAX_HEADER_LEN")
```

x/zkauth/types/zkauth_inputs.go
Mdaiki0730: msk will be 2^bytesPerChunk - 1.
To follow the name bytesPerChunk, `msk := new(big.Int).Sub(new(big.Int).Lsh(big.NewInt(1), uint(bytesPerChunk)*8), big.NewInt(1))`.
In other words, this is not 121 bytes but 121 bits, so I think bitsPerChunk is correct, but am I misunderstanding something?
https://github.com/Finschia/finschia-sdk/pull/1239/files#diff-b762dc12df3343605c06f48c92dc91e6ccaba97a6dfe9d31db88a1ee7f0ca7acR99

zemyblue: Yes, I referred the original source's type, but as you said it's wrong. I'll fix it. Thank you.

Mdaiki0730 (approved):
LGTM

shiki-tak (approved):

PR #1249, opened 2024-02-22
feat: get jwk from keeper

shiki-tak:
## Description

- Save jwks in the keeper field.
- If you want to use jwk for verify, please use GetJWK.
- If there is no corresponding kid when verifying, please run FetchJWK.

## Checklist:


x/zkauth/keeper/keeper.go
```diff
@@ -105,18 +90,61 @@ func (k Keeper) ParseJWKs(byteArray []byte) (jwks []types.JWK, err error) {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(file, resp.Body)
+	if err != nil {
+		return err
+	}
+
+	dataBytes, err := os.ReadFile(targetFile)
```
Mdaiki0730, line 115: How about parsing the resp as is?

zemyblue (commented):
I think we need to do refactoring to fetch class in keeper. 
Because it is a little difficult to fetch jwk in antehandler.

x/zkauth/keeper/keeper.go
shiki-tak: I changed the order of assigning to dataBytes and use WriteFile.

Mdaiki0730: The key is held in both the jwks field and the file, so I don't think it is necessary to save it to the file. What do you think?
https://github.com/Finschia/finschia-sdk/pull/1249/files#diff-2a2241d04eae2d37982aafdf8f50d27ae141f623782c48ee72fbe125ece11fc7R23

shiki-tak: I thought about it again, and it doesn't seem necessary to be stored in both as you say. I stopped saving locally.

PR #1272, opened 2024-03-07
feat: replaces decorators

shiki-tak:
## Description

I made the following three fixes.
1. Activate jwk fetch in simapp and zkauth MsgExecution can be executed.
2. Replacement of the following decorators
- IncrementSequenceDecorator
- SetPubKeyDecorator
- SigVerificationDecorator
3. Other minor fixes

**Note:** Currently, the following antehandler must be commented out in order for simapp that includes the zkauth module to work properly.
https://github.com/Finschia/finschia-sdk/pull/1272/files#diff-68f17637e76320504c315ee15128b4f711050394e820f1c4be80a2e8bdf52c43R43

**Remaining**
1. Around fee
DeductFeeDecorator
SigGasConsumeDecorator

## Checklist:


simapp/app.go
```diff
@@ -109,6 +109,8 @@ import (
 
 	// unnamed import of statik for swagger UI support
 	_ "github.com/Finschia/finschia-sdk/client/docs/statik"
+
+	appante "github.com/Finschia/finschia-sdk/simapp/ante"
```
Mdaiki0730, line 113: It seems that there is no need to separate blocks.

baseapp/block_gas_test.go
```diff
@@ -125,7 +125,7 @@ func TestBaseApp_BlockGas(t *testing.T) {
 				require.Equal(t, []byte("ok"), okValue)
 			}
 			// check block gas is always consumed
-			baseGas := uint64(37352) // baseGas is the gas consumed before tx msg
+			baseGas := uint64(34824) // baseGas is the gas consumed before tx msg
```
Mdaiki0730, line 128: What does this change mean?

x/zkauth/ante/zkauth.go
```diff
@@ -27,25 +31,123 @@ func (zka ZKAuthMsgDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bo
...
+
+				zkisd.ak.SetAccount(ctx, acc)
+			}
+		}
+	}
+
+	return next(ctx, tx, simulate)
+}
+
+func isZKAuthTx(tx sdk.Tx) (bool, []*zkauthtypes.MsgExecution, []types.PubKey, error) {
```
Mdaiki0730, line 130: It appears that isZKAuthTx is responsible for more than just name.
It seems like it can be broken down into three functions for each role.
- isZKAuthTx
- getMsgExecutionFromTx
- getPubkeysFromTx

x/zkauth/ante/zkauth.go
```diff
@@ -27,25 +31,123 @@ func (zka ZKAuthMsgDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bo
...
+		for _, msg := range msgs {
+			for _, signer := range msg.GetSigners() {
+				acc := zkisd.ak.GetAccount(ctx, signer)
+				if err := acc.SetSequence(acc.GetSequence() + 1); err != nil {
+					panic(err)
+				}
+
+				zkisd.ak.SetAccount(ctx, acc)
+			}
+		}
```
Mdaiki0730, line 124: Since the signer of zkMsgs can be obtained directly from zkMsg, how about use it?
```suggestion
		for _, signer := range zkMsg.GetSigners() {
			acc := zkisd.ak.GetAccount(ctx, signer)
			if err := acc.SetSequence(acc.GetSequence() + 1); err != nil {
				panic(err)
			}

			zkisd.ak.SetAccount(ctx, acc)
		}
```

baseapp/block_gas_test.go
shiki-tak: This is a fix for the fact that the gas consumed by test varies depending on the implementation.
This is often arbitrarily changed by bump up etc.

x/zkauth/ante/zkauth.go
shiki-tak: Are the signers of zkMsg EphPubKey? The address I want to set now is for the zkauth address, so I'm trying to get it from the msg signer.

x/zkauth/ante/zkauth.go
shiki-tak: I tried separating it into getMsgExecutionFromTx and getPubkeysFromTx. What do you think?

x/zkauth/ante/zkauth.go
Mdaiki0730: No, they aren't. We can get the zkauth address by zkMsg.GetSigners.
https://github.com/Finschia/finschia-sdk/blob/ca6ca76ea1eb5d3aff74bb8fe8910445c3b456ed/x/zkauth/types/msgs.go#L96-L103

x/zkauth/ante/zkauth.go
Mdaiki0730: I feel strange that the return value of isZKAuthTx includes pubkey and msgExecution.
The previous suggestion meant processing getMsgExecutionFromTx and getPubkeysFromTx after isZKAuthTx, which returns bool or (bool, error).

I think one way is to rename isZKAuthTx.
e.g) getZKAuthInfoFromTx

x/zkauth/ante/zkauth.go
shiki-tak: I understood, I fix it.

x/zkauth/ante/zkauth.go
shiki-tak: getMsgExecutionFromTx and getPubkeysFromTx after isZKAuthTx can certainly be called in that order, but I wanted to only call one function with multiple ante handlers, so I grouped them together.

Mdaiki0730 (approved):

PR #1277, opened 2024-03-12
fix: add init logic of module accounts just in case

jaeseung-bae:
## Description

* (audit issue) add init logic of foundation module accounts to InitGenesis in order to eliminate potential panic
* ~NOTE: This is likely to change account number state in auth module.~ --> InitGenesis never runs more than once

## Checklist:

### DIFF
--- a/x/foundation/keeper/internal/genesis.go
+++ b/x/foundation/keeper/internal/genesis.go
@@ -48,6 +48,9 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data *foundation.GenesisState) erro
 
 	k.SetPool(ctx, data.Pool)
 
+	// init module accounts just in case
+	_ = k.authKeeper.GetModuleAccount(ctx, foundation.ModuleName)
+	_ = k.authKeeper.GetModuleAccount(ctx, foundation.TreasuryName)
 	return nil
 }
 
--- a/x/foundation/keeper/internal/genesis.go
+++ b/x/foundation/keeper/internal/genesis.go
@@ -1,7 +1,10 @@
 package internal
 
 import (
+	"fmt"
+
 	sdk "github.com/Finschia/finschia-sdk/types"
+
 	"github.com/Finschia/finschia-sdk/x/foundation"
 )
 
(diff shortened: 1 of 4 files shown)


ulbqb:
> NOTE: This is likely to change account number state in auth module.

You mean, this may have a negative impact on mainnet?

jaeseung-bae:
> > NOTE: This is likely to change account number state in auth module.
> 
> You mean, this may have a negative impact on mainnet?

A newly joined node that tries to sync with our operating node can never sync, because they may have different account number state.
So i labeled state breaking change, we need to apply this to the next major version.

ulbqb (approved):

x/foundation/keeper/internal/genesis.go
```diff
@@ -48,6 +48,9 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data *foundation.GenesisState) erro
 
 	k.SetPool(ctx, data.Pool)
 
+	// init module accounts just in case
+	_ = k.authKeeper.GetModuleAccount(ctx, foundation.ModuleName)
```
0Tech, line 52: In case of absence of the permission, it will return nil, hence it should check the result, and if necessary, panic.

0Tech (changes requested):

0Tech (approved):

0Tech (dismissed):

0Tech (approved):

x/foundation/keeper/internal/genesis_test.go
```diff
@@ -285,3 +288,58 @@ func TestImportExportGenesis(t *testing.T) {
...
+	gs := &foundation.GenesisState{
+		Params:     foundation.DefaultParams(),
+		Foundation: foundation.DefaultFoundation(),
+	}
+	ctx := app.BaseApp.NewContext(checkTx, tmproto.Header{})
+
+	testCases := map[string]struct {
+		mockAccKeeper *stubAccKeeper
+	}{
+		"failed to generate foundation module account=" + foundation.ModuleName: {
```
0Tech, line 306: You mean, foundation.Treasury? Because the stub contains foundation.ModuleName.

jaeseung-bae: Yes, is it confusing?

0Tech: It says, failed to generate foundation.ModuleName, but it's opposite. Isn't it?

jaeseung-bae: Aha, My mistake.

0Tech: Changing the name of cases, or the logic in the stub would be required.

0Tech (approved):
LGTM

PR #1278, opened 2024-03-13
fix: correct set of address

shiki-tak:
## Description

Modified to assume that iss used when calculating address is received as a base64 encoded value.

## Checklist:


Mdaiki0730:
I didn't realize this at the time of design, but it seems that the appropriate field name is issBase64 instead of issF.
Since it is related to this PR, how about fixing it all at once?
https://github.com/Finschia/finschia-sdk/blob/feat/zkauth/proto/finschia/zkauth/v1beta1/zkauth.proto#L28

Mdaiki0730:
Would you write changelog?

PR #1279, opened 2024-03-14
Deduct gas fee from zkauth address

Mdaiki0730:
## Description

Deduct the gas fee that would have been deducted by DeductFeeDecorator from zkauth address
Mostly the same as DeductFeeDecorator.AnteHandle.
https://github.com/Finschia/finschia-sdk/blob/df06ee16ed85f660b391f69d0732d6b00560c5fa/x/auth/ante/fee.go#L75

In the future there should be a way to branch out with zkauth more generally.

## Checklist:


x/zkauth/ante/zkauth.go
```diff
@@ -152,7 +154,63 @@ func (zdf ZKAuthDeductFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
...
 
-	// Case of zkauth msg, does nothing in this case
+	// Almost no different from the implementation of zdf.dfd.AnteHandle, uses ZKauthAddress for feePayer.
+	feeTx, ok := tx.(sdk.FeeTx)
+	if !ok {
+		return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+	}
+
+	if addr := zdf.ak.GetModuleAddress(authtypes.FeeCollectorName); addr.Empty() {
+		return ctx, fmt.Errorf("fee collector module account (%s) has not been set", authtypes.FeeCollectorName)
```
shiki-tak, line 164: How about using sdkerrors.Wrap like elsewhere?

x/zkauth/ante/zkauth.go
```diff
@@ -152,7 +154,63 @@ func (zdf ZKAuthDeductFeeDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simul
...
+	}
+
+	fee := feeTx.GetFee()
+	signature := zkMsgs[0].GetZkAuthSignature()
```
shiki-tak, line 168: I'm not sure, so I have a question, is it not necessary to run a loop with zkMsgs?Is zkMsgs[0] enough?

shiki-tak (commented):
Please add CHANGELOG

x/zkauth/ante/zkauth.go
Mdaiki0730: The honest answer is that the specifications have not been determined at the moment.
I would like to note in the comments that it is necessary to consider how to respond when multiple zkmsg are included in one tx.

x/zkauth/ante/zkauth.go
Mdaiki0730: This is not an error case for user input.
I can't find a suitable sdkerror, so how about leave it to a higher-level function?

shiki-tak: Surely, or If the result of this function is Empty(), other modules seem to panic.

shiki-tak (dismissed):

x/zkauth/ante/zkauth.go
zemyblue: And we need to check the case that the tx messagea are mixed both zkauth and other normal messages.

zemyblue (approved):
But I agree this pr is good. but we need to check the case that the messages are added multiple and mixing message type. So please add another issue for solving multiple messages.

Mdaiki0730:
> So please add another issue for solving multiple messages.

I added an issue, so I'll merge this PR. Thank you.
https://github.com/Finschia/finschia-sdk/issues/1286

zemyblue (approved):

PR #1280, opened 2024-03-14
feat: fetch jwk considering goroutine exit

shiki-tak:
## Description

- Terminate FetchJWK's goroutine on shutdown.
- fetch_intervals does not need to be included in ProvidersConfig, so I removed it.

## Checklist:


x/zkauth/keeper/keeper.go
```diff
@@ -26,6 +30,8 @@ type Keeper struct {
 
 var _ types.ZKAuthKeeper = &Keeper{}
 
+const fetchIntervals = 60
```
zemyblue, line 33: What are the units? sec or min? Please add it to commend also.

zemyblue (approved):

simapp/app.go
```diff
@@ -372,8 +373,13 @@ func NewSimApp(
...
 
 	// Fetch JWK
+	var wg sync.WaitGroup
+	wg.Add(1)
+
 	ctx := app.BaseApp.NewUncachedContext(true, tmproto.Header{})
-	app.ZKAuthKeeper.FetchJWK(ctx)
+	go app.ZKAuthKeeper.FetchJWK(ctx, &wg)
+
+	wg.Wait()
```
Mdaiki0730, line 382: Doesn't the process stop here?

Mdaiki0730: It seems like I misunderstood, but the process does not stop because it is Done after starting the goroutine with FetchJWT.
By the way, I think the goroutine in FetchJWT will start up even without this wg, what do you think?

Mdaiki0730: `go app.ZKAuthKeeper.FetchJWK(ctx, &wg)`
I mean this code probably doesn't need a goroutine.

x/zkauth/keeper/keeper.go
```diff
@@ -84,40 +91,58 @@ func (k Keeper) DispatchMsgs(ctx sdk.Context, msgs []sdk.Msg) ([][]byte, error)
...
-						k.jwksMap.AddJWK(&jwkCopy)
-					}
-				}
-				time.Sleep(time.Duration(fetchIntervals) * time.Second)
 			}
 		}
 	}()
 }
 
+func (k Keeper) Fetch(ctx sdk.Context) {
```
Mdaiki0730, line 127: Since it is not expected to be used from outside, how about making it private?

simapp/app.go
shiki-tak: Is it the process of block generation? If so, I'm sure it hasn't stopped.

Mdaiki0730: Yes, it doesn't stop. That was my misunderstanding.

> By the way, I think the goroutine in FetchJWT will start up even without this wg, what do you think?
> go app.ZKAuthKeeper.FetchJWK(ctx, &wg)
> I mean this code probably doesn't need a goroutine.

What do you think about this?

shiki-tak: As you said, wg was unnecessary so I deleted it. thank you!

simapp/app.go
```diff
@@ -373,7 +373,7 @@ func NewSimApp(
 
 	// Fetch JWK
 	ctx := app.BaseApp.NewUncachedContext(true, tmproto.Header{})
-	app.ZKAuthKeeper.FetchJWK(ctx)
+	go app.ZKAuthKeeper.FetchJWK(ctx)
```
Mdaiki0730, line 376: It seems that goroutine is also no longer needed.

Mdaiki0730:
```suggestion
	app.ZKAuthKeeper.FetchJWK(ctx)
```

shiki-tak: Oops, thank you!

Mdaiki0730 (approved):

PR #1288, opened 2024-03-19
fix: use accAddress to compare in validatebasic function in collection & token modules

170210:
## Description

This PR uses accAddress instead of address to compare in the `validatebasic` function in the collection & token modules.
Before this PR, it used case-sensitive comparison, thus it would occur a bug when `holder` and `operator` are the same but in different capitalization.

## Checklist:


0Tech (commented):
You may also deal with `MsgRevokeOperator`.

170210:
@0Tech  Do you mean that I should add the same test case to the `TestMsgRevokeOperator`? If so, is it necessary to add the same test case to all test functions?

ulbqb (approved):

0Tech (approved):

PR #1339, opened 2024-04-19
feat: implement fswap module's genesis and params

170210:
## Description

- [x] implement genesis related functions
- [x] implement params & swapped related functions
- [x] add genesis IT & UT tests
- [x] add fswap module into baseApp

## Checklist:


x/fswap/types/params.go
```diff
@@ -1,21 +1,56 @@
...
+	fmt "fmt"
+	"strings"
+
 	"gopkg.in/yaml.v2"
+
+	sdk "github.com/Finschia/finschia-sdk/types"
+)
+
+const (
+	DefaultNewCoinDenom string = "PDT"
```
0Tech, line 14: How about using configuration injected within calling `NewKeeper()`? The application may change the literal without modifying the SDK. This could be also applied to the exchanging rate and the old denom.

x/fswap/keeper/genesis.go
```diff
@@ -0,0 +1,31 @@
...
+	if err := k.SetParams(ctx, genState.Params); err != nil {
+		panic(err)
+	}
+	if err := k.SetSwapped(ctx, genState.Swapped); err != nil {
+		panic(err)
+	}
+	totalOldCoinsSupply := bk.GetSupply(ctx, types.DefaultOldCoins).Amount
+	totalNewCoinsSupply := types.DefaultSwapRate.MulInt(totalOldCoinsSupply)
+	totalNewCoins := sdk.NewDecCoinFromDec(genState.Params.NewCoinDenom, totalNewCoinsSupply)
+	if err := k.SetTotalSupply(ctx, totalNewCoins); err != nil {
```
0Tech, line 20: Is the total supply a read only value? If so, it would be fine, or you may add a comment on the relevant line (on keys.go?).
If not, you should consider exporting it on `ExportGenesis()`.

170210: `Total supply` is a read-only value and does not change once the swap is activated. I will add a comment in this section of the codes and include a line in keys.go to mention this.

x/fswap/types/params.go
170210: Since `DefaultNewCoinDenom` is bound with fswap module, and we will not change it after launch, I have placed it in the params.
About the exchange rate and oldcoindenom, I will inject them as config.

0Tech: Actually, it would be opposite. The argument handed over by `NewKeeper()` would live in the application, and the value itself would be a literal (i.e. constant). The parameter, on the other hand, could be changed by a proposal.
If your "params" means a constant in the SDK, I would also prefer the `NewKeeper()` method, because the application may use a different value in their business logic.

x/fswap/keeper/genesis.go
0Tech: Is it OK to change the value of the total supply after exporting and importing the genesis? We should always consider the scenario.

x/fswap/types/genesis.go
```diff
@@ -1,24 +1,18 @@
...
-
-// DefaultIndex is the default capability global index
-const DefaultIndex uint64 = 1
-
 // DefaultGenesis returns the default Capability genesis state
 func DefaultGenesis() *GenesisState {
 	return &GenesisState{
-		// this line is used by starport scaffolding # genesis/types/default
-		Params: DefaultParams(),
+		Swapped: DefaultSwapped(),
```
0Tech, line 6: The returned value would depend on the literal via `DefaultConfig()`. The default configuration itself would be OK, because the application may modify the config in its initialization. However, the content of `DefaultGenesis()` would NOT be affected by the injected configuration. So you should modify the signature of `DefaultGenesis()`.

0Tech (commented):
We should consider making the swappable amount as a parameter.

x/fswap/types/keys.go
```diff
@@ -17,6 +17,12 @@ const (
...
 
+var (
+	ParamsKey                 = byte(0x00)
+	SwappedKey                = byte(0x01)
+	SwappableNewCoinAmountKey = byte(0x02)
+)
+
 func KeyPrefix(p string) []byte {
 	return []byte(p)
 }
```
jaeseung-bae, line 28: 1. I think, It'd better to move keys to `keeper/keys.go`
Because keeper is responsible of handling states with keys.

2. How about using `collections` util to generate key value in order to update to v0.50 style a little bit easier?
Here is example.
```go
import "cosmossdk.io/collections"

var (
	paramsKeyValue      = collections.NewPrefix(0)
	swappedKeyValue     = collections.NewPrefix(1)
)

func paramsKey() []byte {
	return paramsKeyValue.Bytes()
}
```

170210: About 2, thanks for your suggestion, I will try it.
And about 1, I find in cosmos0.50.0 version, all the keys.go are under types.

170210: About 1, all the keys are under type directory  in cosmos0.50.0 version.
About 2, I tried it before, but failed.

x/fswap/keeper/fswap.go
```diff
@@ -0,0 +1,51 @@
...
+	bz, err := k.cdc.Marshal(&swapped)
+	if err != nil {
+		return err
+	}
+	store.Set([]byte{types.SwappedKey}, bz)
+	return nil
+}
+
+func (k Keeper) GetTotalSupply(ctx sdk.Context) sdk.Coin {
+	store := ctx.KVStore(k.storeKey)
```
jaeseung-bae, line 33: How about this name `GetSwappableNewCoinAmount`?
```suggestion
func (k Keeper) GetSwappableNewCoinAmount(ctx sdk.Context) sdk.Coin {
	store := ctx.KVStore(k.storeKey)
```

x/fswap/types/keys.go
jaeseung-bae: About1, Okay, Thanks for letting me know.
About 2, I also failed to use whole collecions' feature, but was able to use key gen only.

170210: > About 2, I also failed to use whole collecions' feature, but was able to use key gen only.

But how could I use key to set/get, we cannot use `collections.NewItem`

x/fswap/types/genesis.go
```diff
@@ -1,24 +1,28 @@
...
 func (gs GenesisState) Validate() error {
-	// this line is used by starport scaffolding # genesis/types/validate
-
-	return gs.Params.Validate()
+	if err := gs.Params.Validate(); err != nil {
+		return err
+	}
+	if err := gs.Swapped.Validate(); err != nil {
+		return err
+	}
```
0Tech, line 26: The denoms are injected through `NewKeeper()` which allows the application developer to change the denominations. However, this validation logic lacks related verification, which leads to mismatch between the default genesis state and the configuration.
The problem is originated from `Coin` in the state. I recommend you address this issue.

x/fswap/types/errors.go
```diff
@@ -8,5 +8,7 @@ import (
 
 // x/fswap module sentinel errors
 var (
-	ErrSample = sdkerrors.Register(ModuleName, 1100, "sample error")
+	ErrParamsNotFound                 = sdkerrors.Register(ModuleName, 1100, "params does not exist")
+	ErrSwappedNotFound                = sdkerrors.Register(ModuleName, 1101, "swapped does not exist")
+	ErrSwappableNewCoinAmountNotFound = sdkerrors.Register(ModuleName, 1102, "swappable new coin amount does not exist")
 )
```
jaeseung-bae, line 14: I've found any specific reason for this sentinel errors, why don't we use general one `sdkerrors.ErrNotFound`?

170210: I would like to know which part cannot found

x/fswap/types/keys.go
```diff
@@ -17,6 +17,12 @@ const (
 	MemStoreKey = "mem_fswap"
 )
 
+var (
+	ParamsKey                 = byte(0x00)
+	SwappedKey                = byte(0x01)
+	SwappableNewCoinAmountKey = byte(0x02)
```
0Tech, line 23: Can we prune it?

proto/lbm/fswap/v1/query.proto
```diff
@@ -19,7 +19,10 @@ service Query {
 
 message QuerySwappedRequest {}
 message QuerySwappedResponse {
-  Swapped swapped = 1 [(gogoproto.nullable) = false];
+  cosmos.base.v1beta1.Coin old_coin_amount = 1
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/Finschia/finschia-sdk/types.Coin"];
+  cosmos.base.v1beta1.Coin new_coin_amount = 2
+      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/Finschia/finschia-sdk/types.Coin"];
```
jaeseung-bae, line 25: `old_coin_amount` in `message Swapped` has different type to the one in here. 
This is a little bit confusing.

x/fswap/module_simulation.go
```diff
@@ -1 +1,43 @@
...
+	simtypes "github.com/Finschia/finschia-sdk/types/simulation"
+	"github.com/Finschia/finschia-sdk/x/fswap/types"
+)
+
+var _ module.AppModuleSimulation = AppModule{}
+
+// GenerateGenesisState creates a randomized GenState of the module.
+func (AppModule) GenerateGenesisState(simState *module.SimulationState) {
+	fswapGenesis := types.GenesisState{
+		Params:  types.DefaultParams(),
```
0Tech, line 17: The parameter contains the limit, so we should modify it.

proto/lbm/fswap/v1/query.proto
170210: I will change the name in message `EventSwapCoins` and `QueryXXXResponse`
`new/old_coin_amount` will be`types.Int` and `new/old_coin` will be`types.Coin`

x/fswap/types/keys.go
170210: Sorry, I forget to delete `SwappableNewCoinAmount`, `GetSwappableNewCoinAmount` &`SetSwappableNewCoinAmount`

jaeseung-bae (approved):

0Tech (approved):
LGTM

PR #1345, opened 2024-04-26
feat: fswap module's basic functionality(MsgSwap, MsgSwapAll, Query, Proposal)

170210:
## Description

* Add fswap module's basic functionality
  * `Swap`, `SwapAll`, `MakeSwapProposal`

## Checklist:


x/fswap/keeper/keeper.go
```diff
@@ -1,42 +1,345 @@
...
+	}
+	fswapInit, err := k.getFswapInit(ctx)
+	if err != nil {
+		return err
+	}
+	if oldCoinAmount.GetDenom() != fswapInit.GetFromDenom() {
+		return errors.New("denom mismatch")
+	}
+
+	newAmount := oldCoinAmount.Amount.Mul(fswapInit.SwapMultiple)
```
0Tech, line 85: You may check the restriction after the swap has been completed (or in `updateSwapped()`).

x/fswap/keeper/keeper.go
```diff
@@ -1,42 +1,345 @@
...
+		Address:       addr.String(),
+		OldCoinAmount: oldCoinAmount,
+		NewCoinAmount: newCoinAmount,
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (k Keeper) SwapAll(ctx sdk.Context, addr sdk.AccAddress) error {
```
0Tech, line 122: `SwapAll()` now has to get denoms as parameters.

x/fswap/keeper/keeper.go
```diff
@@ -1,42 +1,345 @@
...
+	bz, err := k.cdc.Marshal(&fswapInit)
+	if err != nil {
+		return err
+	}
+	store.Set(fswapInitKey(fswapInit.ToDenom), bz)
+	return nil
+}
+
+func (k Keeper) getAllSwapped(ctx sdk.Context) []types.Swapped {
+	swappedSlice := make([]types.Swapped, 0)
```
0Tech, line 153: May use config for the init.

x/fswap/types/fswap.go
```diff
@@ -21,23 +21,20 @@ func NewSwapped(
 
 // DefaultSwapped returns an initial Swapped object
 func DefaultSwapped() Swapped {
```
0Tech, line 23: Denoms as parameters.

proto/lbm/fswap/v1/tx.proto
```diff
@@ -14,7 +14,7 @@ service Msg {
 message MsgSwapRequest {
```
0Tech, line 14: `MsgSwapRequest` -> `MsgSwap`.
Also, you may consider specifying target denom in the message.

For `MsgSwapAll`, it would be better to provide from-denom in the message.

x/fswap/keeper/keeper.go
```diff
@@ -1,42 +1,341 @@
...
+func (k Keeper) getSwappableNewCoinAmount(ctx sdk.Context) (sdk.Coin, error) {
+	swapCap, err := k.getSwapCap(ctx)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	swapped, err := k.getSwapped(ctx)
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+	denom, err := k.getToDenom(ctx)
```
0Tech, line 236: In `MsgSwap`, the users specify the amount in "from-denom". It would be nice to provide remaining amount in "from-denom".

jaeseung-bae: I thought it's nice to provide remaining amount too. 
But I didn't add, after I checked that `Bank.Send` doesn't return remaining amount after `Send`.

x/fswap/keeper/keeper.go
```diff
@@ -1,42 +1,345 @@
...
-	bk types.BankKeeper,
-	config types.Config,
-) Keeper {
+	AccountKeeper
+	BankKeeper
+
+	fromDenom    string
+	toDenom      string
+	swapMultiple sdk.Int
+	swapCap      sdk.Int
```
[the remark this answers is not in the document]
jaeseung-bae: Updated code will use it. 
thanks for comment

x/fswap/keeper/keeper.go
jaeseung-bae: There are some updated code.
Please check updated code has the same problem.

x/fswap/keeper/keeper.go
jaeseung-bae: There are some updated code.
Please check updated code has the same problem.

proto/lbm/fswap/v1/tx.proto
0Tech: > Is is to name without `Request` suffix cosmos style? Or any other reason?

I know about the grpc convention, but sadly it's the cosmos style. e.g. `/cosmos.bank.v1beta1.MsgSend`.

> For about proto update, do you mean like this?

Also `from_denom` and `to_denom` in `MsgSwapAll`. It's because the state now has generalized structure, so we need to specify the denoms for the conversion.

x/fswap/keeper/keeper.go
0Tech: > I checked that `Bank.Send` doesn't return remaining amount after `Send`.

The bank send does not return remaining amount in the response. And I don't want `MsgSwap` to return the remaining amount either.
My point was, the query returns remaining amount in "to-denom", but in the txs we use amounts in "from-denom". But it seems the behavior is from the requirement. OK to ignore it.

x/fswap/keeper/keeper.go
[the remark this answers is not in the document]
0Tech: We've introduced swap in the state, so I think we don't need (and cannot) use the variable of the keeper for this purpose.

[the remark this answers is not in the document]
jaeseung-bae: Okay, I expected a littble bit faster by using those variables. But this makes code a little bit messy. I'll remove it.

x/fswap/keeper/keeper.go
jaeseung-bae: How about this?
I've added `to_denom` to `MsgSwap` and `from/to_denom` to `MsgSwapAll`.

```protobuf
service Msg {
  rpc Swap(MsgSwap) returns (MsgSwapResponse);
  rpc SwapAll(MsgSwapAll) returns (MsgSwapAllResponse);
}

message MsgSwap {
  // holder's address
  string from_address = 1;
  // from-coin amount
  cosmos.base.v1beta1.Coin from_coin_amount = 2
      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/Finschia/finschia-sdk/types.Coin"];
  string to_denom = 3;
}

message MsgSwapResponse {}

message MsgSwapAll {
  // holder's address
  string from_address = 1;
  string from_denom = 2;
  string to_denom = 3;
}

message MsgSwapAllResponse {}
```

proto/lbm/fswap/v1/tx.proto
jaeseung-bae: How about this?
I've added `to_denom` to `MsgSwap` and `from/to_denom` to `MsgSwapAll`.

```protobuf
service Msg {
  rpc Swap(MsgSwap) returns (MsgSwapResponse);
  rpc SwapAll(MsgSwapAll) returns (MsgSwapAllResponse);
}

message MsgSwap {
  // holder's address
  string from_address = 1;
  // from-coin amount
  cosmos.base.v1beta1.Coin from_coin_amount = 2
      [(gogoproto.nullable) = false, (gogoproto.castrepeated) = "github.com/Finschia/finschia-sdk/types.Coin"];
  string to_denom = 3;
}

message MsgSwapResponse {}

message MsgSwapAll {
  // holder's address
  string from_address = 1;
  string from_denom = 2;
  string to_denom = 3;
}

message MsgSwapAllResponse {}
```

x/fswap/keeper/keeper.go
```diff
@@ -6,37 +6,316 @@ import (
...
+func (k Keeper) swapCap(ctx sdk.Context) (sdk.Int, error) {
+	swapInit, err := k.swapInit(ctx)
+	if err != nil {
+		return sdk.Int{}, err
+	}
+
+	return swapInit.AmountCapForToDenom, nil
+}
+
+func (k Keeper) swapInit(ctx sdk.Context) (types.SwapInit, error) {
```
0Tech, line 271: Now the request `MsgSwap` and `MsgSwapAll` have enough information in its context, we can omit this finder. Applying this, you won't need denom verifications.

x/fswap/keeper/expected_keepers.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	sdk "github.com/Finschia/finschia-sdk/types"
+)
+
+type (
+	AccountKeeper interface {
+		GetModuleAddress(name string) sdk.AccAddress
+	}
+
+	BankKeeper interface {
```
0Tech, line 12: We should consider using `IsSendEnabledCoins()`.

x/fswap/keeper/genesis.go
```diff
@@ -5,21 +5,47 @@ import (
...
+		return types.ErrCanNotHaveMoreSwap.Wrapf("cannot initialize genesis state, there are more than %d swapped", k.config.MaxSwaps)
+	}
+
+	for _, swap := range genState.GetSwaps() {
+		if err := k.MakeSwap(ctx, swap); err != nil {
+			panic(err)
+		}
 	}
+
+	for _, swapped := range genState.GetSwappeds() {
```
0Tech, line 28: It has been already dealt with `ValidateBasic()`.

x/fswap/keeper/genesis.go
```diff
@@ -5,21 +5,47 @@ import (
...
+	if len(genState.GetSwaps()) > k.config.MaxSwaps && !k.isUnlimited() {
+		return types.ErrCanNotHaveMoreSwap.Wrapf("cannot initialize genesis state, there are more than %d swapInits", k.config.MaxSwaps)
+	}
+
+	if len(genState.GetSwappeds()) > k.config.MaxSwaps && !k.isUnlimited() {
+		return types.ErrCanNotHaveMoreSwap.Wrapf("cannot initialize genesis state, there are more than %d swapped", k.config.MaxSwaps)
+	}
+
+	for _, swap := range genState.GetSwaps() {
+		if err := k.MakeSwap(ctx, swap); err != nil {
```
0Tech, line 23: `MakeSwap()` would initialize the corresponding statistics into the default value (zero). We must use dedicated logic (in this case, setxxx variants) in `InitGenesis()`. Also, for the swappeds.

x/fswap/keeper/keeper.go
```diff
@@ -6,37 +6,217 @@ import (
...
+
+func (k Keeper) Swap(ctx sdk.Context, addr sdk.AccAddress, fromCoinAmount sdk.Coin, toDenom string) error {
+	swap, err := k.getSwap(ctx, fromCoinAmount.Denom, toDenom)
+	if err != nil {
+		return err
+	}
+
+	newAmount := fromCoinAmount.Amount.Mul(swap.SwapMultiple)
+	newCoinAmount := sdk.NewCoin(toDenom, newAmount)
+	if err := k.checkSwapCap(ctx, swap, newCoinAmount); err != nil {
```
0Tech, line 48: Both of `checkSwapCap()` and `updateSwapped()` call `getSwapped()` within its logic. It would be nice to refactor relevant flow. e.g. get swapped -> update it -> validate the result -> set swapped.

x/fswap/keeper/proposal.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/Finschia/finschia-sdk/x/fswap/types"
+)
+
+func (k Keeper) MakeSwap(ctx sdk.Context, swap types.Swap) error {
+	if err := swap.ValidateBasic(); err != nil {
+		return err
+	}
+
+	isNewSwap := true
+	if _, err := k.getSwap(ctx, swap.ToDenom, swap.FromDenom); err == nil {
```
0Tech, line 16: Should we have to prohibit the reverse swaps?

x/fswap/types/msgs.go
```diff
@@ -5,64 +5,62 @@ import (
...
 	}
 
-	if !msg.Amount.IsPositive() {
-		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, msg.Amount.String())
+	if !m.FromCoinAmount.IsPositive() {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, m.FromCoinAmount.String())
+	}
+
+	if len(m.GetToDenom()) == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, m.FromCoinAmount.String())
```
0Tech, line 26: It's not a coin, so using another code would be better (invalid request?).

x/fswap/types/msgs.go
```diff
@@ -5,64 +5,62 @@ import (
...
-func (msg MsgSwapAllRequest) ValidateBasic() error {
-	_, err := sdk.AccAddressFromBech32(msg.FromAddress)
+func (m *MsgSwapAll) ValidateBasic() error {
+	_, err := sdk.AccAddressFromBech32(m.FromAddress)
 	if err != nil {
 		return sdkerrors.Wrapf(sdkerrors.ErrInvalidAddress, "Invalid address (%s)", err)
 	}
 
+	if len(m.GetFromDenom()) == 0 {
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidCoins, "Invalid Denom")
```
0Tech, line 51: Specifying "from" denom would be helpful.

x/fswap/keeper/expected_keepers.go
jaeseung-bae: Are you saying that we need to call `IsSendEnabledCoins()` before calling `SendCoinsFromAccountToModule` and `SendCoinsFromModuleToAccount`?
If so, Is it because that two method doesn't check `IsSendEnabledCoins()`?

x/fswap/keeper/proposal.go
jaeseung-bae: There is requirement for one-way swap only.

x/fswap/keeper/expected_keepers.go
0Tech: You don't need to call `IsSendEnabledCoins()` before every single call. Just make sure the denom of user input is enabled for sending (e.g. on message server).

x/fswap/keeper/proposal.go
0Tech: `getSwap()` won't return a valid swap instance for (to, from) tuple. So until the governance explicitly add a swap for (to, from), the swap would be refused. And we can prohibit the number of supported swaps to 1.

x/fswap/keeper/expected_keepers.go
jaeseung-bae: Okay, I'll make a call `IsSendEnabledCoins()` for `Swap` and `SwapAll` txs.

x/fswap/keeper/proposal.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if _, err := k.getSwap(ctx, swap.FromDenom, swap.ToDenom); err == nil {
+		isNewSwap = false
+	}
+
+	if !isNewSwap && !k.config.UpdateAllowed {
+		return errors.ErrInvalidRequest.Wrap("update existing swap not allowed")
+	}
+
+	if err := k.increaseSwapCount(ctx); err != nil {
```
0Tech, line 28: We must increase the count only if `isNewSwap`.

x/fswap/keeper/genesis.go
jaeseung-bae: I fixed InitGenesis()

x/fswap/keeper/proposal.go
```diff
@@ -0,0 +1,121 @@
...
+
+import (
+	"github.com/Finschia/finschia-sdk/store/prefix"
+	sdk "github.com/Finschia/finschia-sdk/types"
+	"github.com/Finschia/finschia-sdk/types/errors"
+	"github.com/Finschia/finschia-sdk/x/fswap/types"
+)
+
+func (k Keeper) MakeSwap(ctx sdk.Context, swap types.Swap) error {
+	if err := swap.ValidateBasic(); err != nil {
```
0Tech, line 11: It has been already checked within `ValidateBasic()` of the proposal.

jaeseung-bae:
WIP:
- Add `DenomMetadata` to `MakeSwapProposal`
- Add `QuerySwaps`

x/fswap/keeper/proposal.go
jaeseung-bae: Thanks, I'll remove it

x/fswap/keeper/proposal.go
```diff
@@ -43,6 +44,8 @@ func (k Keeper) MakeSwap(ctx sdk.Context, swap types.Swap) error {
 	if err != nil {
 		return err
 	}
+
+	k.SetDenomMetaData(ctx, toDenomMetadata)
```
0Tech, line 48: We may prevent modifying any existing denom metadata.

jaeseung-bae: What conditions can be used?
Should we allow `SetDenomMetaData` only for new `Swap`?

x/fswap/keeper/keeper_test.go
```diff
@@ -165,7 +169,8 @@ func (s *KeeperTestSuite) TestSwap() {
 			s.Require().NoError(err)
 
 			actualAmount := s.keeper.GetBalance(ctx, tc.from, s.swap.GetToDenom()).Amount
-			expectedAmount := tc.expectedBalanceWithoutMultiply.Mul(s.swap.SwapMultiple)
+			multipliedAmountDec := s.swap.SwapRate.Mul(sdk.NewDecFromBigInt(tc.expectedBalanceWithoutMultiply.BigInt()))
```
0Tech, line 172: The most critical logic in the swap would be the multiplication. It would be nice to provide the expected value in the cases as a literal.

x/fswap/keeper/proposal.go
0Tech: There would be a possibility to override existing denom for a new swap (e.g. set `to_denom` to cony in finschia application). You may get denom metadata of the proposal, and if it already exists, compare the values. It should fail if the denom metadata differs from the existing one.

jaeseung-bae: Should we prevent existing metadata when it is different to new one?

0Tech: It depends. If we want to open a way to modify existing metadatas, it would not be required. If we just want to "add" a new metadata, we have to compare the values.

jaeseung-bae: >How about accepting metadata only for the first time?
When I think about it again, there would be more swap with the same to-denom..

x/fswap/keeper/keeper_test.go
jaeseung-bae: I extract calculation logic and fix a bug.
Please review again.

x/fswap/keeper/keeper.go
```diff
@@ -83,6 +82,12 @@ func (k Keeper) Swap(ctx sdk.Context, addr sdk.AccAddress, fromCoinAmount sdk.Co
 	return nil
 }
 
+func (k Keeper) CalcSwap(swapRate sdk.Dec, fromCoinAmount sdk.Int) sdk.Int {
```
0Tech, line 85: It does not need to be a method of the keeper.

x/fswap/keeper/keeper.go
```diff
@@ -83,6 +82,12 @@ func (k Keeper) Swap(ctx sdk.Context, addr sdk.AccAddress, fromCoinAmount sdk.Co
...
+	multipliedAmountDec := swapRate.Mul(sdk.NewDecFromBigInt(fromCoinAmount.BigInt())).TruncateInt()
```
0Tech, line 86: Why not `return swapRate.MulTruncate(sdk.NewDecFromInt(fromCoinAmount)).TruncateInt()`

0Tech (dismissed):

proto/lbm/fswap/v1/fswap.proto
```diff
@@ -4,9 +4,43 @@ package lbm.fswap.v1;
...
+      [(gogoproto.customtype) = "github.com/Finschia/finschia-sdk/types.Int", (gogoproto.nullable) = false];
+  string swap_rate = 4
+      [(gogoproto.customtype) = "github.com/Finschia/finschia-sdk/types.Dec", (gogoproto.nullable) = false];
+}
+
+message SwapStats {
+  option (gogoproto.equal)            = true;
+  option (gogoproto.goproto_stringer) = false;
+
+  int32 swap_count = 1;
```
zemyblue, line 26: Is this the count of all swaps regardless of the type of swap registered?

jaeseung-bae: This counts registered swaps only.

x/fswap/keeper/proposal.go
```diff
@@ -0,0 +1,161 @@
...
+	bank "github.com/Finschia/finschia-sdk/x/bank/types"
+	"github.com/Finschia/finschia-sdk/x/fswap/types"
+)
+
+func (k Keeper) MakeSwap(ctx sdk.Context, swap types.Swap, toDenomMetadata bank.Metadata) error {
+	stats, err := k.getSwapStats(ctx)
+	if err != nil {
+		return err
+	}
+	if int(stats.SwapCount) >= k.config.MaxSwaps && !k.isUnlimited() {
```
0Tech, line 16: It must be `>` instead of `>=`. And this verification must be relocated to the back. It's because we should allow the updates.

0Tech: You must also relocate the verification. If not, it will allow one to surpass the threshold.

0Tech (approved):

0Tech (approved):

x/fswap/types/codec.go
```diff
@@ -2,22 +2,48 @@ package types
...
+	govtypes "github.com/Finschia/finschia-sdk/x/gov/types"
 )
 
 func RegisterCodec(cdc *codec.LegacyAmino) {
 	// this line is used by starport scaffolding # 2
 }
 
-func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
+func RegisterLegacyAminoCodec(cdc *codec.LegacyAmino) {
+	cdc.RegisterConcrete(&MakeSwapProposal{}, "finschia-sdk/MakeSwapProposal", nil)
```
zemyblue, line 20: There is no legacy.RegisterAminoMsg for `MsgSwap` and `MsgSwapAll`. Please check it.

x/fswap/types/codec.go
```diff
@@ -2,22 +2,48 @@ package types
...
+		(*govtypes.Content)(nil),
+		&MakeSwapProposal{},
+	)
 }
 
 var (
-	Amino     = codec.NewLegacyAmino()
-	ModuleCdc = codec.NewProtoCodec(cdctypes.NewInterfaceRegistry())
+	amino     = codec.NewLegacyAmino()
+	ModuleCdc = codec.NewProtoCodec(types.NewInterfaceRegistry())
```
zemyblue, line 35: I think the `ModuleCdc` should be AminoCodec. 
```
ModuleCdc = codec.NewAminoCodec(amino)
```

zemyblue (approved):

0Tech (approved):

PR #1365, opened 2024-05-08
fix: update swap keys for possibly overlapped keys

jaeseung-bae:
## Description

* fix: update swap keys for possibly overlapped keys
  * They should be the same -> `(hello, world)` != `(hell, oworld)`

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,7 @@ Ref: https://keepachangelog.com/en/1.0.0/
 * (x/mint, x/slashing) [\#1323](https://github.com/Finschia/finschia-sdk/pull/1323) add missing nil check for params validation
 * (x/server) [\#1337](https://github.com/Finschia/finschia-sdk/pull/1337) fix panic when defining minimum gas config as `100stake;100uatom`. Use a `,` delimiter instead of `;`. Fixes the server config getter to use the correct delimiter (backport cosmos/cosmos-sdk#18537)
 * (x/fbridge) [\#1361](https://github.com/Finschia/finschia-sdk/pull/1361) Fixes fbridge auth checking bug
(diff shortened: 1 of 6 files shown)


0Tech (commented):
We should prohibit denoms longer than 255.

jaeseung-bae:
> We should prohibit denoms longer than 255.

Can i have any reason for 255 limit?

0Tech:
> > We should prohibit denoms longer than 255.
> 
> Can i have any reason for 255 limit?

It's because our newly introduced lengthPrefix() assumes it (converts int to byte).

jaeseung-bae:
> > > We should prohibit denoms longer than 255.
> > 
> > 
> > Can i have any reason for 255 limit?
> 
> It's because our newly introduced lengthPrefix() assumes it (converts int to byte).

Thanks for letting me know.

I found regex for denom validation `[a-zA-Z][a-zA-Z0-9/-]{2,127}`.
function to get key will be called after `validation logic`, so it's okay without length limit.

zemyblue (approved):

jaeseung-bae:
> > > > We should prohibit denoms longer than 255.
> > > 
> > > 
> > > Can i have any reason for 255 limit?
> > 
> > 
> > It's because our newly introduced lengthPrefix() assumes it (converts int to byte).
> 
> Thanks for letting me know.
> 
> I found regex for denom validation `[a-zA-Z][a-zA-Z0-9/-]{2,127}`. function to get key will be called after `validation logic`, so it's okay without length limit.

I got another case to call without `validation logic`.
I've just made commit to fix it.

0Tech (approved):

zemyblue (approved):

PR #1387, opened 2024-05-17
feat: add new query to get a swap

jaeseung-bae:
## Description

* Add new query to get a swap
* Add missing swagger fswap entry for generation
  * Add new URL for fswap's query json to `client/docs/config.json`
* Fix for missing pagination for `QuerySwapsRequest` CLI

## Checklist:


x/fswap/keeper/grpc_query_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	app             *simapp.SimApp
+	ctx             sdk.Context
+	queryClient     types.QueryClient
+	keeper          keeper.Keeper
+	swap            types.Swap
+	toDenomMetadata bank.Metadata
+	fromDenom       string
+	toDenom         string
+}
```
170210, line 34: Why not use `KeeperTestSuite` in keeper_test, it looks like same. And other modules like bank module also use same test Class in keeper.

jaeseung-bae: Okay, I'll check. I just wanted test-suite for grpc_query.
Since I don't expect more testcase with different pre-conditions, I'll check keeper's test-suite to use.

jaeseung-bae: After I checked, I think I'll keep `FSwapQueryTestSuite` for grpc_query test.
Their test-cases need a little different pre-condition and It's okay to have their own test-suite.

170210: But it seems `KeeperTestSuite` and `FSwapQueryTestSuite` does the same things, isn't it?
Both of them set a dummy initial swap data.

jaeseung-bae: No, They are different.
Test suite for Query doesn't need creation of accounts with initial balance and the test suite only requires a registered `Swap` which `KeeperTestSuite` does not.

x/fswap/types/msgs.go
```diff
@@ -141,3 +144,20 @@ func (m *MsgSetSwap) Route() string {
 func (m *MsgSetSwap) GetSignBytes() []byte {
 	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(m))
 }
+
+func (m *QuerySwapRequest) Validate() error {
+	if m == nil {
+		return status.Error(codes.InvalidArgument, codes.InvalidArgument.String())
+	}
+	if m.GetFromDenom() == m.GetToDenom() {
+		return status.Error(codes.InvalidArgument, codes.InvalidArgument.String())
```
0Tech, line 153: It would be nice to provide the specific reason.

0Tech (commented):
+1 for describing the swagger fix in the PR description.
It would be also nice to give a hint about the fix regarding the missing pagination handling.

x/fswap/types/msgs.go
jaeseung-bae: I added descriptive messages for invalid `QuerySwapRequest`

0Tech (dismissed):

zemyblue (commented):
There is no single swap query cli. Please check it.

jaeseung-bae:
> There is no single swap query cli. Please check it.

Thanks for finding missing cli. I've just added.

zemyblue (approved):

0Tech (approved):

PR #1391, opened 2024-05-21
test: add cli_test for fswap module

170210:
## Description

closes: #XXXX

## Checklist:


x/fswap/client/testutil/tx.go
```diff
@@ -0,0 +1,379 @@
...
+				s.toDenom.Base,
+				fmt.Sprintf("--%s=true", flags.FlagSkipConfirmation),
+				fmt.Sprintf("--%s=%s", flags.FlagBroadcastMode, flags.BroadcastBlock),
+				fmt.Sprintf("--%s=%s", flags.FlagFees, sdk.NewCoins(sdk.NewCoin(s.cfg.BondDenom, sdk.NewInt(1))).String()),
+			},
+			true,
+			0,
+			nil,
+		},
+		{
```
0Tech, line 86: Is this case relevant to the coverage of x/fswap?

x/fswap/client/cli/tx.go
```diff
@@ -22,6 +22,10 @@ const (
 	FlagSwapRate            = "swap-rate"
 )
 
+type ToDenomMeta struct {
+	Metadata bank.Metadata `json:"metadata"`
+}
+
```
jaeseung-bae, line 28: In general, It's not desirable to make it public for test.
Is there any way to test without exposing it?

170210: Do you have some ideas?

x/fswap/client/testutil/tx.go
170210: I want to simulate a situation that `coin amount` is invalid. Or I change it into `1abc1`?

x/fswap/client/cli/tx.go
jaeseung-bae: Please wait, let me check.

jaeseung-bae: First you need to revert `ToDenomMeta` to private, then you can change `TestMsgSetSwap` code like below.

```go
func (s *IntegrationTestSuite) TestMsgSetSwap() {
	val := s.network.Validators[0]
	clientCtx := val.ClientCtx
	// avoid printing as yaml from CLI command
	clientCtx.OutputFormat = jsonOutputFormat

	denomMeta := struct {
		Metadata bank.Metadata `json:"metadata"`
	}{
		Metadata: s.toDenom,
	}
```

jaeseung-bae: Using the same tag(`json:"metadata"`) is enough to run this test.

170210: Yes, but this way defining the same structure in two places indeed increases the maintenance cost, as any changes to the structure need to be manually synchronized across both definitions. Is this acceptable?

jaeseung-bae: I can't come up another solution for this but exposing code for testing is a lot worse.

x/fswap/client/testutil/tx.go
0Tech: I mean, the case for "not enough fees" seems like redundant.

0Tech: Also for "not enough gas".

x/fswap/client/cli/tx.go
0Tech: I also have same opinion. Exposing implementation detail as public API is usually bad idea. Because, after exposing it, we cannot change them or the change must be treated as API breaking change.

x/fswap/client/testutil/tx.go
170210: These two tests are referenced by the bank module tests. I think they also can be deleted, I will modify it.

0Tech (approved):

jaeseung-bae (approved):

PR #1393, opened 2024-05-23
build: add suffix string to docker container

jaeseung-bae:
## Description

* ~refresh docker container's bound volume which refer to current working directory~
* use current directory as a suffix for docker container to run it per their environment
  * ~Without refreshing docker container with bound volume location,~ It doesn't generate `proto`, `swagger` depending on the current working directory
  * For example
    * Does work for `~/Finschia/fisnchia-sdk`
    * Doesn't work for `~/xxx/finschia-sdk`
      * to work this, you should remove docker container first.

## Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -394,8 +394,9 @@ easyjson-gen:
 
 containerProtoVer=v0.2
 containerProtoImage=tendermintdev/sdk-proto-gen:$(containerProtoVer)
-containerProtoGen=cosmos-sdk-proto-gen-$(containerProtoVer)
-containerProtoGenSwagger=cosmos-sdk-proto-gen-swagger-$(containerProtoVer)
+containerSuffix=$(shell echo $(CURDIR) | sed 's/\//./g')
+containerProtoGen=cosmos-sdk-proto-gen-$(containerProtoVer)-$(containerSuffix)
+containerProtoGenSwagger=cosmos-sdk-proto-gen-swagger-$(containerProtoVer)-$(containerSuffix)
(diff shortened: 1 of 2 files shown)


tkxkd0159 (commented):
Then, it should be deleted if the containerProtoGen already exist instead of starting. Because this change assumes that the container will be regenerated every execution moment.

0Tech (commented):
Or, we may just append the project name to the name of the container, which prevents the reuse by other projects.

jaeseung-bae:
> Then, it should be deleted if the containerProtoGen already exist instead of starting.

Yes, It is. It should be deleted.

> Or, we may just append the project name to the name of the container, which prevents the reuse by other projects.

Thanks for the comment, I'll try this one.

Makefile
```diff
@@ -394,8 +394,9 @@ easyjson-gen:
 
 containerProtoVer=v0.2
 containerProtoImage=tendermintdev/sdk-proto-gen:$(containerProtoVer)
-containerProtoGen=cosmos-sdk-proto-gen-$(containerProtoVer)
-containerProtoGenSwagger=cosmos-sdk-proto-gen-swagger-$(containerProtoVer)
+containerSuffix=$(shell echo $(CURDIR) | sed 's/\//./g')
```
0Tech, line 397: The path may contain bad characters (e.g. space), so it would be better to replace them. e.g. `sed 's/[^[:alnum:]_-]/./g'`

0Tech (approved):

zemyblue (approved):

PR #1395, opened 2024-05-23
fix: return error instead of panic for behaviors triggered by client

tkxkd0159:
## Description

closes: #XXXX

## Checklist:


x/fbridge/keeper/transfer.go
```diff
@@ -23,7 +23,7 @@ func (k Keeper) handleBridgeTransfer(ctx sdk.Context, sender sdk.AccAddress, amo
 	}
 
 	if err := k.bankKeeper.SendCoinsFromAccountToModule(ctx, sender, types.ModuleName, token); err != nil {
-		panic(err)
+		return 0, err
 	}
 
 	if err := k.bankKeeper.BurnCoins(ctx, types.ModuleName, token); err != nil {
```
jaeseung-bae, line 29: How about changing this case as well?

From:
```go
if err := k.bankKeeper.BurnCoins(ctx, types.ModuleName, token); err != nil {
    panic(fmt.Errorf("cannot burn coins after a successful send to a module account: %v", err))
```
To: 
> you can use %w to wrap another error
```go
if err := k.bankKeeper.BurnCoins(ctx, types.ModuleName, token); err != nil {
    return 0, fmt.Errorf("cannot burn coins after a successful send to a module account: %w", err)
```

0Tech (commented):
It could be nice to address other panics on x/fbridge. I found 2 points:

- critical
    - keeper/abci.go#L36: In theory, `updateRole()` can return error by registering two identical role suggestions. So we have to update the error handling or the chain halts.
- good-to-have
    - keeper/genesis.go#L22 and L28: `setRole()` and `setBridgeSwitch()` may return error because of its argument verification. It that case, `InitGenesis()` panics, instead of returning errors. However, if `InitGenesis()` returns any errors, eventually it will panic by the module. Hence, it would be good to return errors or modify the signature of `InitGenesis()` so it won't return any errors.

x/fbridge/keeper/transfer.go
tkxkd0159: this case must not be executed and cannot be triggered by client. So I think it's better to keep it.

x/fbridge/keeper/auth_test.go
```diff
@@ -52,10 +52,6 @@ func TestAssignRole(t *testing.T) {
 	err = k.updateRole(ctx, types.RoleOperator, addrs[1])
 	require.NoError(t, err)
 	require.Equal(t, types.RoleMetadata{Guardian: 1, Operator: 1, Judge: 0}, k.GetRoleMetadata(ctx))
-
```
0Tech, line 55: Or, by confirming it does not return an error, we can recycle the obsolete test logic into a nice verification on the role update specification (i.e. it will silently ignore the suggestion, making it idempotent).

x/fbridge/keeper/auth.go
```diff
@@ -61,10 +61,6 @@ func (k Keeper) addVote(ctx sdk.Context, proposalID uint64, voter sdk.AccAddress
 
 func (k Keeper) updateRole(ctx sdk.Context, role types.Role, addr sdk.AccAddress) error {
 	previousRole := k.GetRole(ctx, addr)
-	if previousRole == role {
```
0Tech, line 64: We may return at this point.

0Tech (commented):
And please mind keeper/genesis.go#L22.

x/fbridge/keeper/transfer.go
jaeseung-bae: Okay.

0Tech (approved):

x/fbridge/keeper/abci.go
```diff
@@ -32,11 +34,17 @@ func (k Keeper) EndBlocker(ctx sdk.Context) {
 		}
 
 		if types.CheckTrustLevelThreshold(k.GetRoleMetadata(ctx).Guardian, voteYes, guardianTrustLevel) || proposal.Proposer == k.GetAuthority() {
-			if err := k.updateRole(ctx, proposal.Role, sdk.MustAccAddressFromBech32(proposal.Target)); err != nil {
```
0Tech, line 35: Now, we can revert this change thanks to idempotent behavior of `updateRole()`.

0Tech (approved):

jaeseung-bae (approved):

PR #1415, opened 2024-06-11
chore: add additional testcases for fswap

jaeseung-bae:
## Description

* add more tests for fswap

## Checklist:


x/fswap/keeper/grpc_query_test.go
```diff
@@ -42,17 +43,17 @@ func (s *FSwapQueryTestSuite) SetupTest() {
...
 	s.keeper = s.app.FswapKeeper
 
-	pebSwapRateForCony, err := sdk.NewDecFromStr("148079656000000")
+	kaiSwapRateForCony, err := sdk.NewDecFromStr("148079656000000")
 	s.Require().NoError(err)
 	swapCap := sdk.NewInt(1000)
 	s.Require().NoError(err)
 	s.fromDenom = "cony"
-	s.toDenom = "kei"
+	s.toDenom = "kai"
```
zemyblue, line 51: `kei` is right.

x/fswap/keeper/genesis_test.go
```diff
@@ -45,3 +51,78 @@ func (s *KeeperTestSuite) TestInitAndExportGenesis() {
...
+						Amount: sdk.ZeroInt(),
+					},
+					ToCoinAmount: sdk.Coin{
+						Denom:  "bbb",
+						Amount: sdk.ZeroInt(),
+					},
+				},
+			},
+		}
+	}
```
170210, line 83: Could this function as a private function of `genesis_test` and also used in `TestInitAndExportGenesis()`

x/fswap/keeper/grpc_query_test.go
```diff
@@ -88,48 +89,229 @@ func (s *FSwapQueryTestSuite) SetupTest() {
...
+		},
+		{
+			name: "invalid request",
+			request: &types.QuerySwapsRequest{
+				Pagination: &query.PageRequest{
+					Key:        []byte("invalid-key"),
+					Offset:     1,
+					Limit:      0,
+					CountTotal: false,
+					Reverse:    false,
```
170210, line 304: I think these two lines could be deleted. (Other test functions which test pagination haven't these options)

x/fswap/types/msgs_test.go
```diff
@@ -89,49 +92,328 @@ func TestAminoJSON(t *testing.T) {
 func TestQuerySwapRequestValidate(t *testing.T) {
```
170210, line 92: I think these part is duplicated in `msg_server_test`.
Could we standardize the validation of messages? Except for the swap, the other three messages do not have a validation function. I suggest that QuerySwapRequest could also be validated similarly. And currently it only checks if the value is null without using sdk.ValidateDenom, which I think is insufficient.

jaeseung-bae: After I checked duplication, `TestQuerySwapRequestValidate` is duplicated and redundant.
Thank you, I'll remove it

> And currently it only checks if the value is null without using sdk.ValidateDenom, which I think is insufficient.
I agree with it. 
I want to add test cases only and don't want to change production code. We may update insufficient validation logic in other PR.

zemyblue (approved):

170210 (approved):

